PR #7, opened 2017-09-03
CMake improvements

jschueller:
Hi,
I hope these can be included in the official repo repository as I the site doesn't explain thow to contribute to the hg repository.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,4 @@
 # TODO:
-# 1. pkgconfig file not yet built.
 # 2. installation not tested.
 # 3. Problems generating html pages.
 # 4. create usable libtool .la file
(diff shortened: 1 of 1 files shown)


0-wiz-0:
Thanks for the patch.
The standard way is to send patches to the mailing list.

I looked at the diff. Instead of hardcoding a different man page installation path, I switched to using GNUInstallDirs, now you can set it from the command line:
-DCMAKE_INSTALL_MANDIR:PATH=share/man

The pkgconfig part is incomplete. @LIBS@ can contain other libraries as well, in particular the bzip2 one.  Can you please fix that?

jschueller:
ok, I updated the pkgconfig to include "-lbz2" and removed the man dir change

0-wiz-0:
I've modified and merged the patch, see https://github.com/nih-at/libzip/commit/4fba7af60267214d752510b27b9d612effe5fa25
The includedir was wrong, and I added some quoting to make it more like the `autoconf` generated one.
Thank you for your contribution!

PR #10, opened 2017-09-15
add zip_get_version  and zip_get_version_id functions 

remicollet:
to retrieve runtime version (which may differ from buildtime version when ABI is compatible)

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -161,6 +161,7 @@ SET(LIBZIP_SOURCES
   zip_unchange_archive.c
   zip_unchange_data.c
   zip_utf-8.c
+  zip_version.c
 )
 
 IF(WIN32)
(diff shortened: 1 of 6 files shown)


0-wiz-0:
We have added a `zip_libzip_version()` function that returns the version string.
We don't really understand the need or point of the other functions, can you please explain your use case?

remicollet:
> We don't really understand the need or point of the other functions,

Bah... just something quite commonly used in other libraries.

BTW zip_libzip_version is enough for my needs.

remicollet:
BTW use case

```
if (zip_get_version_id(0) >= 0x130) {
  // use BZ2 compression 
} else {
  // use other compression
}
```

0-wiz-0:
That use case does not work: bzip2 support is a compile-time option, so you can have a new libzip without bzip2 support.

PR #18, opened 2017-12-18
Don't use `unused` in aescrypt macro

carl-mastrangelo:
The word `unused` interferes with gcc macros like `__attribute__((unused))`.


0-wiz-0:
Merged, thank you!

PR #32, opened 2018-03-05
Appveyor and assorted fixes

janisozaur:


janisozaur:
I'm sure some corners could be cut here to reduce time for each run, but right now it works and tests all the configurations vcpkg has to offer, including {ARM,x86,x86-64}-UWP and native (non-UWP) ARM with MSVC.

One low-hanging fruit is caching the downloaded cmake zip, it takes around 1m30s in each job!

dillof:
Thank you for getting Appveyor to work. It's greatly appreciated. Also, thanks for the assorted Windows fixes.

janisozaur:
@dillof many thanks for merging this! I hope this will ensure MSVC compatibility for all your future releases.

@IntelOrca can you verify if the symbols are no longer exposed in static builds?

@dillof once that is confirmed, would you mind doing a point-release so I can update vcpkg, as noted here: https://github.com/Microsoft/vcpkg/pull/2937 ?

IntelOrca:
@janisozaur 👍 It is resolved now.

janisozaur:
@dillof @0-wiz-0 could you please do a point release with those fixes? I'd like to submit update to vcpkg and then use this updated version in OpenRCT2.

Thanks in advance.

0-wiz-0:
I've just made a 1.5.0 release with your contributions. Thank you, @janisozaur!

ras0219-msft:
Update is pushed to Vcpkg 🛩 Thanks @janisozaur @0-wiz-0!

PR #35, opened 2018-03-19
Fix build with LibreSSL

heirecka:
LibreSSL unfortunately defines "OPENSSL_VERSION_NUMBER 0x20000000L"
but doesn't support the OpenSSL 1.1 API.


dillof:
Thanks, slightly modified and applied:
I've factored out the API version check so it's only in one place.

PR #41, opened 2018-04-05
Don't set RPATH on installed libraries on Linux

darkxst:
Linux distributions generally don't set RPATH's on installed binaries, Ubuntu/Debian in particular have a pretty strong policy on this. The only time you should need to set RPATH on linux is when you link against private libraries in your own project.

I Don't know enough about the other platforms you support, as to what their policies on RPATH are, so I have set this change to only affect Linux.


praiskup:
On Thursday, April 5, 2018 12:19:49 PM CEST Thomas Klausner wrote:
> Is this really something that can be decided for "Linux" in general, or are
> there Linux distributions that do like rpath?  (I'm thinking of adding a
> switch to turn off rpath instead.)

Speaking for Fedora/RHEL/CentOS, yes - rpath is not allowed **iff** we
speak about installation of libraries into standard path (/usr/lib +
ldconf paths).  I would say that *by default* this is not loved feature.

But if one decided to ship newer build/version of DSO (than the one
provided by default on particular system), one needs to install the
library on on-standard path so linker doesn't see it by default.  In such
cases, **opt-in** rpath support is useful for **dependant** projects.
Third-party distributors are the target audience here though, rather than
the distributions themselves.  And speaking for libzip.so, it doesn't seem
to depend on that exotic set of libraries so this shouldn't be needed.

PR #42, opened 2018-04-05
Fix syntax errors in man pages

darkxst:
As picked up by lintian and quotes were not rendered in man pages

```
W: libzip-dev: manpage-has-errors-from-man usr/share/man/man3/zip_libzip_version.3.gz 21: warning: can't find special character `Lq'
W: ziptool: manpage-has-errors-from-man usr/share/man/man1/ziptool.1.gz 328: warning: can't find special character `Lq'
```

### DIFF
--- a/man/zip_libzip_version.man
+++ b/man/zip_libzip_version.man
@@ -18,7 +18,7 @@ libzip (-lzip)
 .SH "DESCRIPTION"
 \fBzip_libzip_version\fR
 returns the version number of the library as string in the format
-\(Lq$MAJOR.$MINOR.$MICRO$SUFFIX\(Rq
+\(lq$MAJOR.$MINOR.$MICRO$SUFFIX\(rq
 where
 \fI$MAJOR\fR
 is the major version,
--- a/man/ziptool.man
+++ b/man/ziptool.man
@@ -325,9 +325,9 @@ Add a file called
 to the zip archive
 \fItestbuffer.zip\fR
 with data
-\(LqThis is a test.\en\(Rq
+\(lqThis is a test.\en\(rq
 where
-\(Lq\en\(Rq
+\(lq\en\(rq
 is replaced with a newline character:
 .nf
 .sp


ischwarze:
Hi Tim,

thanks to Thomas Klausner (wiz at NetBSD) for forwarding this bug report to me.

Of course, the patch is invalid and should not be merged:  You are patching an auto-generated file.

But as usual, bad patches trigger good ones.  I just fixed the root cause upstream at the mandoc project:

http://mandoc.bsd.lv/cgi-bin/cvsweb/mdoc_man.c#rev1.124

Thanks again for the report,
Ingo

P.S.
Unrelated to this bug report, but the Debian package *really* ought to install the high-quality *.mdoc versions of the manual pages instead of the stunted *.man versions.  The man(7) versions are merely provided for prehistoric systems like Oracle Solaris 11 and the like.  Modern systems like Debian support mdoc(7) natively and perfectly, no matter whether the user installs groff (which is the Debian default) or mandoc as the manual page formatter.

0-wiz-0:
Fixed, thanks for the bug report!

darkxst:
Hey guys,
  Thanks for the quick fix, I clearly didn't realise they were autogenerated ;) 

I am working on hopefully sneaking the libzip 1.5.0 update into Ubuntu 18.04 ([LP Bug](https://bugs.launchpad.net/bugs/1674057). And will forward my packaging work to the Debian Maintainer at some point.

I can look into making sure the packaging uses the mdoc files, but my understanding is that it is the upstream buildsystem that installs the man pages into a virtual tree as "mdoc.7" etc, then the debhelper tools split them out into other packages if required. Not sure the Debian tools ever see the mdoc files?

@ischwarze: Does that sound right you? 

Essentially is it libzip CMake system needs to installs the mdoc ? or does it actually need overrides in the acutal Debian packaging

darkxst:
@0-wiz-0 thanks! Please also look at my RPATH patch, that is a major blocker in Ubuntu/Debian, although I have just distro patched it for now ;)

PR #48, opened 2018-06-14
Update CMakeLists.txt

SleepySpaceDad:
Added TARGET_INCLUDE_DIRECTORIES() for automatic inclusion of the libzip headers when linking with TARGET_LINK_LIBRARIES()


0-wiz-0:
I read the documentation for `TARGET_INCLUDE_DIRECTORIES()`, but that didn't make it clearer to me why this might be needed here.
What problem are you seeing? How does this fix the problem?
Thanks.

SleepySpaceDad:
Yeah, let me clarify. That's my bad. I'm new at this.
So, using target_include_directories() allows the user to only have to link with libzip using target_link_libraries() and the headers are already included. It's better than relying on include_directories() which can potentially make hidden dependencies in large projects using libzip.

johnb003:
@SleepySpaceDad He's right, this should be done, but it's incomplete as is.

This needs to remove the regular "INCLUDE_DIRECTORIES" that it effectively replaces.
And it should be like:

TARGET_INCLUDE_DIRECTORIES(zip
  PUBLIC:
        $<INSTALL_INTERFACE:include>
        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/..>
)

These generator expressions state:
PUBLIC: ie (building the target itself -- that's private scope, and anyone that links the target gets these properties too -- that's public scope)

For building this target look for headers in the current source dir, and in the parent dir.
And if anything is doing add_subdirectory to the lib folder and finding this target, they can simply add "zip" to their target_link_libraries, and they'll automatically also get those source folders for finding headers without having to state include dirs anywhere explicitly.

This will make more sense if my PR #50 is approved which adds a target export.

With the exported target, you can find_package(zip), and then just target_link_library(Foo zip::zip)

and wherever the package is installed to, "<installed to path>/include"
The install path by default on a linux machine is like, /usr/local/<something> I think, though, this can be overwritten with: CMAKE_INSTALL_PATH when building.

Further, lets say you exported everything to a local folder on your machine, if done correctly, the cmake target installed with the package can all be relocated to another machine even (Assuming it's the same architecture, and can find_package the same dependencies), it will work there as well, using CMAKE_PREFIX_PATH to indicate where things are installed.

johnb003:
I realize I was a bit ambiguous: "This will make more sense if my PR is approved."

I meant, this change will have more value, in combination with my PR.
Not that, my PR contains it in a better way.

That said, since you closed it, it does now. ;)

PR #49, opened 2018-06-19
Update zip_source_win32w.html

redlasha:
The parameter appears to be invalid.


0-wiz-0:
Thank you! The PR is for a generated file, so I committed the changes to the source file instead and regenerated the file.

Issue #59, opened 2018-08-07
feature request: option to not build the tools


goutnet:
Hi,

would you accept a pull request to have the option not to build the tools, but only the library?

I need this to integrate properly libzip in an ndk project. The problem being that new versions of the ndk do not provide stdout/stderr (needed by zipmerge and other tools).

Regards,

0-wiz-0:
Merged, thanks!

PR #60, opened 2018-08-07
adding customization for build

goutnet:
This patch adds 4 options to allow custom build of libzip.

```
OPTION(BUILD_TOOLS "Build tools in the src directory (zipcmp, zipmerge, ziptool)" ON)
OPTION(BUILD_REGRESS "Build regression tests" ON)
OPTION(BUILD_EXAMPLES "Build examples" ON)
OPTION(BUILD_DOC "Build documentation" ON)
```

This allow easier integration in projects when all is not needed (or would be impossible to build ex: android NDK).


goutnet:
related issue #59

PR #62, opened 2018-08-29
Adding option to disable SO versionning (for android build)

goutnet:
For the purpose of integrating libzip in an android project, the built .so file must show SONAME versioning (as android ld loader does not support that feature).

For this reason, I added an option `SHARED_LIB_VERSIONNING` allowing this versioning to be removed.

The option is ON by default (which means that versioning is added by default, just like before), but can be turned off by `-DSHARED_LIB_VERSIONNING=OFF`


0-wiz-0:
The patch is clear enough.
I don't understand the explanation though, what does this mean?

> For the purpose of integrating libzip in an android project, the built .so file must show SONAME versioning (as android ld loader does not support that feature).

goutnet:
When using libzip as a dependency of another lib (say `liba.so` depends on `libzip.so`), under Android, the Android linker needs the depends to be on `libzip.so` not `libzip.so.5` (for instance), as the android dynamic linker does not support this feature (basically, if the depends points to `libzip.so.5`, the library will not load, as only `.so` suffixes are supported).

I integrate libzip as a dependency of an android project, (included as .so file), and for this reason, I need the version name to not be added as SONAME in the generated .so file.

This feature can be useful for cross-compiling with NDK  or to other platform that do not support advanced ld library loaders.

I hope that clarifies.

goutnet:
When using libzip as a dependency of another lib (say `liba.so` depends on `libzip.so`), under Android, the Android linker needs the depends to be on `libzip.so` not `libzip.so.5` (for instance), as the android dynamic linker does not support this feature (basically, if the depends points to `libzip.so.5`, the library will not load, as only `.so` suffixes are supported).

I integrate libzip as a dependency of an android project, (included as .so file), and for this reason, I need the version name to not be added as SONAME in the generated .so file.

This feature can be useful for cross-compiling with NDK  or to other platform that do not support advanced ld library loaders.

I hope that clarifies.

0-wiz-0:
Thanks for the explanation. It's weird, but it's easy to support; I've just merged your pull request.

PR #63, opened 2018-09-07
Add an option to disable bzip2 support

vmiklos:
The explicit switch is still on by default, but this way it's possible
to disable bzip2 support even if bzip2 is available on the
system.

This is useful if bzip2 is not needed, but would automatically lead to
an unwanted dependency.


vmiklos:
Thanks for the fast review. :-)

PR #67, opened 2018-10-03
ESP platform compatibility

dexterbg:
This includes the changes necessary to enable building & using libzip on the ESP-32 platform (Espressif ESP-IDF).

Note: I haven't found CMake rules for mbed TLS, so I need to leave this for some CMake guru or add this later on. To build for ESP-32, you currently need to use the include files from here:
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/tree/master/vehicle/OVMS.V3/components/zip

Regards,
Michael


dillof:
We just committed a variant of your stack usage reduction that keeps the buffers on the stack by default, but adds a define you can set to allocate them on the heap (ZIP_ALLOCATE_BUFFER).

lib/zip_source_filep.c
```diff
@@ -253,14 +252,20 @@ create_temp_output(struct read_file *ctx) {
     }
     sprintf(temp, "%s.XXXXXX", ctx->fname);
 
-    mask = umask(_SAFE_MASK);
+#ifdef HAVE_UMASK
+    mode_t mask = umask(_SAFE_MASK);
```
0-wiz-0: Please keep variable definitions at the beginning of blocks.

lib/zip_source_filep.c
```diff
@@ -412,8 +417,6 @@ read_file(void *state, void *data, zip_uint64_t len, zip_source_cmd_t cmd) {
 #endif
 
     case ZIP_SOURCE_COMMIT_WRITE: {
-	mode_t mask;
```
0-wiz-0: Same here -- keep it in this location, please.

lib/zip_source_filep.c
```diff
@@ -423,10 +426,12 @@ read_file(void *state, void *data, zip_uint64_t len, zip_source_cmd_t cmd) {
 	    zip_error_set(&ctx->error, ZIP_ER_RENAME, errno);
 	    return -1;
 	}
-	mask = umask(022);
+#if defined(HAVE_UMASK) && defined(HAVE_CHMOD)
```
0-wiz-0: Neither of these are currently defined. They need to be added to the top level `CMakeLists.txt` and `cmake-config.h.in`.
Also, the Windows case where `_chmod` and `_umask` exist needs to be handled -- there we add `define`s for chmod and umask (see `lib/compat.h`).

0-wiz-0:
Changes requested, but diff not updated, and the diff does not even apply cleanly any more.
Please reopen or make a new PR when the issues are addressed.
Thank you!

PR #71, opened 2018-10-31
Remove inconsistent zip_source_free for zero-length files

### LINKED ISSUE #70: Inconsistent error handling in _zip_open
Specifically, if there is an error with _zip_allocate_new in the zero length file case, the source is freed. In all other cases, care is taken to ensure that source remains valid.

shahms:
Fixes #70


shahms:
I don't have write access to merge.

PR #86, opened 2019-03-27
cmake: option to skip creating dist target

malytomas:
I want to use the library as a cmake subproject (add_subdirectory) in a larger project. Unfortunately, some other library also defines target dist. Dist is, in fact, so common name, that protecting it with variable or `if(TARGET dist)` has become common practice. I would like you to consider merging this as it would reduce maintenance overhead on my and possibly others side. Thanks


0-wiz-0:
This
````
IF(NOT SKIP_DIST)
````
looks at least incorrectly named.
It should either be
````
IF (SKIP_DIST)
````
or a working variation of
````
IF (NOT TARGET dist)
````
Can you please modify your patch?

malytomas:
I apologize, I may have confused you with my incorrect comment. It should be `IF(NOT TARGET dist)` or `IF(NOT SKIP_DIST)`.
By default, the variable SKIP_DIST is not defined and the target should be created (to keep original behavior).
I believe the code is correct.
Should I change the name anyway? Perhaps `IF(NOT NO_DIST)` or `IF(NOT DISABLE_DIST)`?

dillof:
` IF(NOT TARGET dist)` seems like the best solution. If you embed libzip in another project, our dist target isn’t useful anyway.

malytomas:
I have modified the patch.

0-wiz-0:
I've just merged it, thank you!

PR #88, opened 2019-03-31
libmbedcrypto is used directly, so link to it.

kensington:
Otherwise, build can fail with a strict compiler.


0-wiz-0:
Thanks! Actually we don't need the tls version of the library at all. I've fixed this.

PR #97, opened 2019-05-03
Fix pkg-config Libs.private for static linking

lachs0r:
Holy shit, CMake is BAD. And this is a really ugly hack for a thing that would have been trivial in *any* other build system. Please switch to Meson. I’m even willing to help you port it, just make this horrible abomination go away.


0-wiz-0:
Can you please explain the problem you're solving?

(We're not planing to move away from CMake currently, it works well enough.
And meson does not even properly handle rpath yet.)

lachs0r:
Static linking in a MinGW cross toolchain setup. Without this change, there are libraries missing from Libs.private (e.g. bcrypt), so projects using pkg-config will fail to link libzip statically. This change derives the list directly from the target instead of building it separately.

Akaricchi:
What is wrong with Meson's handling of rpath?

0-wiz-0:
@Akaricchi see e.g. https://github.com/mesonbuild/meson/issues/4027

PR #102, opened 2019-05-22
Hunterize libzip

rbsheth:
This PR adds the appropriate CMake changes to make libzip compatible with the Hunter package manager (https://github.com/ruslo/hunter).

If `HUNTER_ENABLED` is `OFF`, the original/default CMake behavior is preserved.

If `HUNTER_ENABLED` is `ON`, libzip will now use Hunter to pull OpenSSL, BZip2, and ZLIB.


0-wiz-0:
Thank you for the pull request. We have discussed this and decided we're currently not interested in merging this.
It makes the CMake files quite a bit larger; we don't have the infrastructure/setup to test this or keep it up-to-date; we hadn't heard of hunter before; and we were told it's already quite easy to include libzip as part of another CMake-based project.

PR #106, opened 2019-06-09
Unix: Fix build when non of the crypto options is available

tsdgeos:
On 89f457f19b7829305e089da41356966d9fa64f88 _zip_mkstempm started using
zip_random

zip_random was previously only used in crypto related places that were
dissabled if no crypto option was found, but since now it's used
unconditionally we need the zip_random_unix.c fallback.


tsdgeos:
P.S: I ha ve a feeling that something similar is needed for windows since there's an existing zip_random_win32.c file that doesn't seem to be used but i don't have access to any windows machine so that's on you

dillof:
Thanks for the heads up, I've been working on a more comprehensive fix that also handles failures more gracefully, which I've just committed.

(Windows doesn't use zip_mkstempm(), so it shouldn't be a problem there.)

PR #119, opened 2019-08-12
Add fuzz target to build

randy408:
Add fuzz target build to regression tests to prevent bitrot.


randy408:
Looks like travis CI doesn't ship clang with libFuzzer and the version check is failing, will have to rethink this..

randy408:
Added a C entry point instead, this doesn't require a C++ compiler nor libFuzzer.

randy408:
Any comments? @0-wiz-0 @dillof

0-wiz-0:
Can you please explain the point of the new executable, and how it will be used by what? Thanks.

randy408:
The new executable is built as a part of the regression tests, it acts as a build test for the fuzz target `zip_read_fuzzer.cc`.
It takes a file path, reads the file and passes it to `LLVMFuzzerTestOneInput()`, which is what `zip_read_fuzzer.cc` would do if it was built with a recent version of clang++ with libFuzzer support.  A C-only standalone driver is a simpler solution that that's also used by other projects e.g. [sqlite](https://www.sqlite.org/src/artifact/d9f1a6f43e7bab45).

randy408:
`fuzz_main.c` [includes](https://github.com/nih-at/libzip/pull/119/files#diff-868e9d80ac25cbc96abf9efcd2d9172dR1) `zip_read_fuzzer.cc` to keep it simple, adding a `.cc` file as source might confuse CMake into requiring a c++ compiler.

0-wiz-0:
Thanks for the explanations!

PR #120, opened 2019-08-13
Fix up Xcode compile warnings and errors using Xcode 10.3 zip_close.c…

prniii:
…, zip_filerange_crc.c, and zip_source_window.c has unreachable code that should be bracketed by #ifdef ZIP_ALLOCATE_BUFFER.

Xcode projects need updating, -Wno-nullability-extension to quiet the clang warning that nullable is a clang extension, remove -L/opt/pgk/lib  prevents compiling in the general case where using Apple CommonCrypto.
All projects except in-memory example compile without warnings, except for Xcode's new localization settings.   in-memory project still fails to compile.

Have been working at integrating libzip in a  project on MacOS and Windows, ran into compiler warnings and errors building on MacOS using Xcode build project, fixed most of them here.

### DIFF
--- a/lib/zip_close.c
+++ b/lib/zip_close.c
@@ -524,10 +524,12 @@ copy_data(zip_t *za, zip_uint64_t len) {
     size_t n;
     double total = (double)len;
 
+#ifdef ZIP_ALLOCATE_BUFFER
     if (!byte_array_init(buf, BUFSIZE)) {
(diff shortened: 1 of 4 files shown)


dillof:
This Xcode project is not meant to compile production builds. I've made this clearer in the repository.

As for the unused code warnings: we'll try to come up with a less intrusive solution.

PR #122, opened 2019-08-25
Use documented CMake way of linking zlib

mmha:
[`ZLIB_LIBRARY` is undocumented and internal to CMake.](https://cmake.org/cmake/help/v3.0/module/FindZLIB.html) Use `ZLIB_LIBRARIES` instead.

Also replace the homebrew version check with the built-in version check from `find_package`.

These changes are necessary to build with current versions of CMake.


0-wiz-0:
Thanks!

PR #127, opened 2019-10-07
fix tests for windows

malytomas:
I have fixed most issues with the perl test runner to make it work on windows.
It is far from perfect, there are few hardcoded paths etc, but it does its job.
The test with piping file from stdin is currently broken, I do not know how to fix it.
The other failing tests might be actual failures in the code. The test empty_2 confirms my prejudice.
I hope it did not break anything on linux, we shall see.
I would be delighted to know your thoughts on this.


dillof:
Thanks, this looks very promising, thank you.

We can fix the remaining tests after merging your changes, but to do so, we would need to get the output of the failing tests. Can you set the environment variable VERBOSE to yes for the ctest invocation?

dillof:
Thanks, I see the details of the test failures now. We’ll fix the easy ones and then merge your change.

We really appreciate your contribution, it helps us keep libzip working properly on Windows.

malytomas:
Great.

I expect a follow-up PR that will fix the search paths for the programs.
I think it will change the _api_ of the perl script to pass in the test configuration (debug/relwithdebinfo/etc.), which is why I postpone it into a separate PR.

Issue #131, opened 2019-10-17
Use of deprecated form of CMake functions prevents use in modern CMake


leecbaker:
```
The plain signature for target_link_libraries has already been used with
  the target "zip".  All uses of target_link_libraries with a target must be
  either all-keyword or all-plain.

  The uses of the plain signature are here:

   * lib/libzip/lib/CMakeLists.txt:223 (TARGET_LINK_LIBRARIES)
```

Libzip uses the old form of the `target_link_libraries()` command which does not have the `PUBLIC`/`PRIVATE`/`INTERFACE` keywords used to specify how this link dependency should be used.

When used as part of CMake code using current versions of the `target_link_library()` and other functions, `PUBLIC`/`PRIVATE`/`INTERFACE` keywords must be specified. This is encountered e.g. when libzip is used as a library in a bigger project.

Instructions on how to appropriately use these keywords is here: https://cmake.org/cmake/help/v3.0/manual/cmake-buildsystem.7.html#build-specification-and-usage-requirements

**libzip Version**
master

**Operating System and Compiler**
Any version of CMake from at least 3.0 and newer; perhaps back as old as CMake 2.8. I observed this with CMake 3.12.

**Solution**
Add the `PUBLIC`/`PRIVATE`/`INTERFACE` keywords to each use of `target_link_libraries()` `target_include_directories()` and other functions, such as [here](https://github.com/nih-at/libzip/blob/master/lib/CMakeLists.txt#L228).

0-wiz-0:
I'm actually not quite clear on the difference between the three types. Could you open a pull request with a suggested fix, please?

leecbaker:
Yes, no problem.

0-wiz-0:
We marked libzip's use of libraries as `PRIVATE` in 38b2ceea.
I don't see the point in doing the same for executables, if you need that as well in your environment, please explain in more detail.
Thanks!

PR #132, opened 2019-10-17
CMake: add scope to target_link_libraries()

leecbaker:
This is necessary when using libzip as a submodule and using a newer version of CMake.

Fixes https://github.com/nih-at/libzip/issues/131


lib/CMakeLists.txt
```diff
@@ -225,7 +225,7 @@ SET_TARGET_PROPERTIES(zip PROPERTIES VERSION 5.0 SOVERSION 5)
 ENDIF()
 
 
-TARGET_LINK_LIBRARIES(zip ${ZLIB_LIBRARIES} ${OPTIONAL_LIBRARY})
+TARGET_LINK_LIBRARIES(zip PUBLIC ${ZLIB_LIBRARIES} ${OPTIONAL_LIBRARY})
```
0-wiz-0: Just for clarification: Are the (compression) libraries that libzip depends on really `PUBLIC`? None of their symbols are passed on to the outside via header files.

leecbaker: As I understand it, PUBLIC mirrors the previous behaviour.

If symbols from the other libraries aren't visible from outside, then it sounds like `PRIVATE` might be the more correct way to go, but I'm not familiar enough with how this works to say for sure.

Flamefire: > As I understand it, PUBLIC mirrors the previous behaviour.

Correct. But judging from the usage those libraries are only internally used, so they can be PRIVATE. Simple rule of thumb: If a dependent libraries headers are included in this libraries "public headers" (expected to be used by users), then use PUBLIC, else PRIVATE.

leecbaker:
I've made a couple of changes; the zlib headers should no longer be included in the interface to the libzip library. It's ready for review again, if anybody wants to take a look.

0-wiz-0:
Addressed #131 differently, thanks!

PR #133, opened 2019-10-20
Let zip.h be in a non system include directory

s0m30n3:
I'm using libzip in my project but for clarity, I moved include files to a dedicated libzip folder so I could do #include <libzip/zip.h>. Unfortunately, it fails because of <> style inclusion hence my patch.

If you find it useful, please merge.
Thanks.

### DIFF
--- a/lib/zip.h
+++ b/lib/zip.h
@@ -42,7 +42,7 @@ extern "C" {
 #endif
 #endif
 
-#include <zipconf.h>
+#include "zipconf.h"
 
 #ifndef ZIP_EXTERN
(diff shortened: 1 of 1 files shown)


dillof:
That is not a supported configuration.

Also, installed headers should include other installed headers with <>; "" is for user headers. If we include zipconf.h with "", then a zipconf.h present in the current directory would interfere with correct compilation.

I'm afraid you will have to patch your copy of zip.h, and I would recommend changing it to `#include <libzip/zipconf.h>'.

PR #134, opened 2019-10-21
prevent test runner from searching for files in directories for other build types

malytomas:
The previous PR had few hardcoded paths that were used when searching for programs.
It could incorrectly pick a file from directory for other build type.
This PR replaces the hardcoded paths with paths derived from the build type.

Running tests will require providing the build type name now. See the travis file for an example.
(Maybe someone more clever can improve the makefiles?)


0-wiz-0:
We fixed this slightly differently -- you can pass a sub directory to search in to `runtest`; this way the diff was much smaller. I hope this works as intended!

PR #136, opened 2019-11-03
Add ENABLE_CRYPTO to allow disabling crypto.

Symbitic:
Embedded targets for example, might wish to forego the use of cryptographic libraries intentionally, so they probably don't want a warning message.


0-wiz-0:
What's the problem with disabling the separate supported libraries with the existing flags?

Symbitic:
> What's the problem with disabling the separate supported libraries with the existing flags?

It sends a warning if all encryption libraries are disabled.

Ordinarily, that's probably what you want. You want the user to know they are compiling libzip without the ability to decode encrypted zip files.

But in other circumstances (embedded systems, only a specific set of zip features will be used, etc) you don't want a CMake warning, because you specifically do not want encryption support added.

dillof:
Is this warning a serious problem for you, or just an aesthetic annoyance?

We already have too many build options, so I'm reluctant to add another one.

Symbitic:
Well I suppose you could just change the warning to a message.

No CMake warning is ever a "serious problem" as long as it still builds. But in my case, libzip is being built using ExternalProject, so the warning gets called quite often.

If you think it would be more than what it's worth though, I can live without it.

PR #141, opened 2019-12-13
Removing FILE_ATTRIBUTE_TEMPORARY before Moving the tmp file under Win…

tschabo:
FILE_ATTRIBUTE_TEMPORARY has to be removed after moving the the temporary ZIP file to the right place.


lib/zip_source_win32w.c
```diff
@@ -136,6 +136,13 @@ _win32_rename_temp_w(_zip_source_win32_read_file_t *ctx)
 {
     if (!MoveFileExW(ctx->tmpname, ctx->fname, MOVEFILE_REPLACE_EXISTING))
 	return -1;
+    DWORD attributes = GetFileAttributesW(ctx->fname);
+    if (!attributes)
+    return -1;
+    if (FILE_ATTRIBUTE_TEMPORARY | attributes){
```
0-wiz-0: Should this be
```
if (FILE_ATTRIBUTE_TEMPORARY & attributes){
```
instead?

lib/zip_source_win32w.c
```diff
@@ -136,6 +136,13 @@ _win32_rename_temp_w(_zip_source_win32_read_file_t *ctx)
...
+    if (!SetFileAttributesW(ctx->fname, attributes - FILE_ATTRIBUTE_TEMPORARY))
```
0-wiz-0: We usually do
````
attributes & ~FILE_ATTRIBUTE_TEMPORARY
````

File: lib/zip_source_win32w.c [hunk 3]
```diff
@@ -136,6 +136,13 @@ _win32_rename_temp_w(_zip_source_win32_read_file_t *ctx)
 {
     if (!MoveFileExW(ctx->tmpname, ctx->fname, MOVEFILE_REPLACE_EXISTING))
 	return -1;
+    DWORD attributes = GetFileAttributesW(ctx->fname);
+    if (!attributes)
+    return -1;
```
0-wiz-0: Please use similar indentation as the indentation of the rest of the code.

lib/zip_source_win32w.c
tschabo: Shame on me!
You are right.

lib/zip_source_win32a.c
```diff
@@ -111,7 +111,17 @@ static int
 _win32_rename_temp_a(_zip_source_win32_read_file_t *ctx)
 {
     if (!MoveFileExA(ctx->tmpname, ctx->fname, MOVEFILE_REPLACE_EXISTING))
-	return -1;
+        return -1;
+
+    DWORD attributes = GetFileAttributesA(ctx->fname);
+    if (!attributes)
```
0-wiz-0: Thanks for the fixes. One more thing: `GetFileAttributesA` and `GetFileAttributesW` return `INVALID_FILE_ATTRIBUTES` on errors (see https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfileattributesa) so the check should compare against that, and not zero. Please fix this, then I'll merge.

0-wiz-0:
Thanks for the contribution and the fixes! I've just merged it.

Issue #98, opened 2019-05-04
libzip-1.5 fails to compile on linux with musl-libc with error:zipcmp.c:(.text+0xd74): undefined reference to `fts_read' and 'fts_close'


stefson:
the last version known to work is libzip-1.3

this is the full build.log [build.log.gz](https://github.com/nih-at/libzip/files/3144415/build.log.gz)

I tried both clang-7 and gcc-8.2.0, to no avail.

dillof:
The problem is that fts.h is found, but fts_open() et. al. are not present. Could you please check the man page for fts_open, wether it needs additional libraries?

stefson:
The problem is, that musl doesn't have an own fts implementation. Turns out I had a copy installed which was borrowed from netbsd, removing that and the compile of libzip is working again. Your config scripts detects fts as missing, and does something magic it seems :-)

PR #143, opened 2019-12-17
cmake: detect fts also in external lib

micmac1:
Not all libc implementations support fts. fts support may be added via
an external library.

One such example is musl libc. It doesn't support fts. But support for
fts can be added to the target device via musl-fts, which provides
libfts.

When libzip encounters such a setup (fts.h available but symbols in
libfts.so) the build fails at the linking stage:

zipcmp.c:(.text.startup+0x130): undefined reference to `fts_open'
/home/sk/tmp/openwrt/staging_dir/toolchain-mips_24kc_gcc-8.3.0_musl/lib/gcc/mips-openwrt-linux-musl/8.3.0/../../../../mips-openwrt-linux-musl/bin/ld: zipcmp.c:(.text.startup+0x172): undefined reference to `fts_read'

This commit sets up cmake so that it checks for fts symbols first in
libc and - if none encountered there - in an external libfts. So in case
libc provides fts nothing changes, libzip will use libc's fts symbols
(if fts.h is present). If on the other hand libc doesn't provide fts but
libfts does, it'll link it in where required (zipcmp).

Signed-off-by: Sebastian Kemper <sebastian_ml@gmx.net>


micmac1:
Hi all,

We ran into the same issue described in issue #98 on the OpenWrt build bots. I've proposed this patch as a remedy. Kindly review it when you get around to it.

Thanks!
Seb

micmac1:
Ping @mhei

0-wiz-0:
Thank you for the contribution! I've just merged it.

PR #144, opened 2019-12-18
cmake: improve fts variable handling

micmac1:
HAVE_LIB_FTS is used unconditionally, but there are paths through the
previous logic that result in this variable not being set (example:
HAVE_FTS_H is FALSE). This is not a big issue, but it's not nice either.

This commit adds ELSE() to set HAVE_LIB_FTS also for these paths.

Finally, this commit also sets up FTS_LIB as a proper cache variable
with type INTERNAL, to be in line with CHECK_FUNCTION_EXISTS and
CHECK_LIBRARY_EXISTS, which as well set cache variables of the same
type.

Tested with:

1. musl without musl-fts
2. musl with musl-fts
3. uclibc (includes fts support without extra libfts, like glibc)

Signed-off-by: Sebastian Kemper <sebastian_ml@gmx.net>


micmac1:
Hi all,

It doomed on me this morning that my PR was lacking in the thoroughness department. Sorry for this, I should have gotten it right the first time.

Kind regards,
Seb

Issue #61, opened 2018-08-24
PKWARE encryption encode support


hgbutte:
For backward compatibility reasons i need to create zip files with PKWARE style encryption.
Tried AES-128 but my counterpart does not support this.

PR #153, opened 2020-03-02
add encryption with Traditional PKWARE

eraow:
## Description
Just like #61, we would also like to encrypt zip with Traditional PKWARE for backward compatibility.

## Solution
This implementation mainly complies with AES encryption/decryption implementation and refactors PKWARE decryption to some extent.
I would like to explain 2 points about this PR.

### 1. add CRC calculation before encryption header creation.

https://github.com/eraow/libzip/blob/87d1ab468d4cab91e1a303edc81fcbe4a6e5fc5d/lib/zip_close.c#L456-L463

I think this process is necessary because of the 3 reasons below.
* [CRC value is used at encryption header](https://github.com/nih-at/libzip/blob/master/docs/appnote.txt#L2429).
* [Encryption header is stored at the start of the data area](https://github.com/nih-at/libzip/blob/master/docs/appnote.txt#L2366). (Therefore, encryption header should be created when opening or starting to read the target file)
* [Current CRC layer](https://github.com/eraow/libzip/blob/87d1ab468d4cab91e1a303edc81fcbe4a6e5fc5d/lib/zip_close.c#L415) can not finish the calculation until reading the last of the data.

(In case of AES, precedent MAC calculation is not needed.
 This is because MAC immediately follows [the last byte of encrypted data](https://github.com/nih-at/libzip/blob/master/docs/aes_info.txt#L319) and MAC creation can be finished [at the end of the reading](https://github.com/eraow/libzip/blob/9bd28f945c383b9c014ac851d932ef3cb10c009f/lib/zip_source_winzip_aes_encode.c#L166).)

### 2. use mtime as encryption header last value when Bit 3 is set.

https://github.com/eraow/libzip/blob/87d1ab468d4cab91e1a303edc81fcbe4a6e5fc5d/lib/zip_source_pkware_encode.c#L153-L157

[According to appnote.txt](https://github.com/eraow/libzip/blob/master/docs/appnote.txt#L750), CRC is not used in local file header when Bit 3 is set.
However, appnote.txt seems not to remark what must be used as encryption header value in such a case.
Therefore, I followed libzip pkware decryptor and other products(zipinfo, zipclock, windows zip decryptor, and so on) behavior.


0-wiz-0:
Thank you very much for this pull request!

We noticed that as implemented, it would read all of the data twice, once for computing the CRC and once for actually encrypting it.

Since some sources might not support that, we changed the implementation to use the mtime method that InfoZIP also uses (and which we tried out to be also supported on MacOS and Windows by the builtin implementations).

PR #156, opened 2020-03-08
Fix mingw builds

janisozaur:
1. `srandom` and `random` are POSIX extensions not present on non-POSIX
   platforms, replace them with `srand` and `rand` respectively. They
   _might_ produce different results, but that shouldn't matter in the
   context of the caller, which is the non-secure version
2. Add sources providing mkstemp function to libzip libraries

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -164,6 +164,8 @@ SET(LIBZIP_SOURCES
 
 IF(WIN32)
   SET(LIBZIP_OPSYS_FILES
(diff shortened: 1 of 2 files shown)


0-wiz-0:
Thanks for the pull request! I enhanced it for the UWP version.

janisozaur:
You didn't fix the problem, though: https://github.com/nih-at/libzip/commit/2bba8fcea39c446d77e4671d550931417d9a4fbf#r37699033

PR #163, opened 2020-04-03
fix compiler warning for double macro definition

farindk:
I got this compiler warning when using libzip on Android and including `inttypes.h` before `zip.h`.
This patch fixes it.


0-wiz-0:
Hm. I am not sure about the proper fix for this.
If you include `inttypes.h` before `zip.h` and before defining `#define __STDC_FORMAT_MACROS 1`, does the right thing happen?
Perhaps we need to document that you should include `inttypes.h` (or the other headers) after `zip.h`

farindk:
Requiring a specific include file order does not sound like a clean solution to me. Especially since they can be pulled in through long, unknown chains in system includes that are not under developer control.

Like in my case, where `inttypes.h` is pulled in probably through `<string>` and defines `__STD_FORMAT_MACROS` itself (see below).

I don't see any problem with my proposed patch. It is not defining more macros like before. It is just checking not to define the same macro twice.

````
../android-jni-libs/zip/out/armeabi-v7a/include/zipconf.h:21:9: warning: '__STDC_FORMAT_MACROS' macro redefined [-Wmacro-redefined]
#define __STDC_FORMAT_MACROS 1
        ^
../android-sdk/sdk/ndk/21.0.6113669/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1/inttypes.h:248:12: note: previous definition is here
#   define __STDC_FORMAT_MACROS
````

farindk:
BTW: in the system headers, it is solved in exactly the same way. Excerpt from 'inttypes.h':

````
#if defined(__cplusplus) && !defined(__STDC_FORMAT_MACROS)
#   define __STDC_FORMAT_MACROS
#endif
````

0-wiz-0:
I'm not convinced this is correct. But it seems lots of other code does the same.
If it causes any problems, I'll back it out and you'll have to live with the warning.

PR #175, opened 2020-05-18
CMake patches for Windows

OgreTransporter:
The documentation https://cmake.org/cmake/help/latest/variable/CMAKE_COMMAND.html says that it is better to use `${CMAKE_COMMAND} -E` than `cmake -E`, so the lines https://github.com/nih-at/libzip/blob/223c57e0d4ae0f93b7f9f839d6196be3981d6f0f/CMakeLists.txt#L106 and https://github.com/nih-at/libzip/blob/223c57e0d4ae0f93b7f9f839d6196be3981d6f0f/CMakeLists.txt#L110 would have to be adjusted. In addition, the whole block should be excluded under Windows, because most of the commands do not work. The lines https://github.com/nih-at/libzip/blob/223c57e0d4ae0f93b7f9f839d6196be3981d6f0f/regress/CMakeLists.txt#L52 and https://github.com/nih-at/libzip/blob/223c57e0d4ae0f93b7f9f839d6196be3981d6f0f/regress/CMakeLists.txt#L53 should also be adjusted.

In line https://github.com/nih-at/libzip/blob/223c57e0d4ae0f93b7f9f839d6196be3981d6f0f/lib/CMakeLists.txt#L196 the next problem occurs. SH is not available under Windows. The target doesn't make much sense either, because the file zip_err_str.c is even checked in. I would either delete the target and maintain the file manually or delete the target and the file and have the file generated by CMake.


0-wiz-0:
I didn't want to disable the `dist` target on Windows completely - it is for making releases mostly and noone else should need to use it, but it might work (in MingW) so I don't think it should be disabled.

Thank you for your other changes!

OgreTransporter:
Ok, then just replace `NOT WIN32` by `NOT MSVC`.

PR #190, opened 2020-07-13
restore GNUInstallDirs

remicollet:
Needed to install in proper PATH, (especially  using "lib64" on RPM distro)


0-wiz-0:
Can you please explain in detail what improvements this change brings for you?
Thanks.

remicollet:
On RPM distro (Fedora, RHEL, CentOS...) we build using

 ```
/usr/bin/cmake \
        -DINCLUDE_INSTALL_DIR:PATH=/usr/include \
        -DLIB_INSTALL_DIR:PATH=/usr/lib64 \
        -DSYSCONF_INSTALL_DIR:PATH=/etc \
        -DSHARE_INSTALL_PREFIX:PATH=/usr/share \
        -DLIB_SUFFIX=64 \
   .....

```

So we expect library to be installed in /usr/lib64, not in /usr/lib

GNUInstallDirs is a common/best practice.

So this is not a improvement, rather a regression fix

Issue #180, opened 2020-06-06
make cmake install optional


malytomas:
Please add an option into cmake to disable install.

I am using libzip via add_subdirectory and I do not want libzip to be installed alongside my superproject files. I build libzip as a static library and do not need it in the installation.

Thanks.

0-wiz-0:
Sure.
Is there some CMake support for this (some standard macros or common patterns) or do I just add a `LIBZIP_DO_INSTALL` variable that defaults to `ON` and make installation dependent on that?

malytomas:
I am not aware of any common way to do this (unfortunately).
`LIBZIP_DO_INSTALL` sounds good.

0-wiz-0:
Implemented, thanks for the suggestion.

PR #203, opened 2020-08-07
fix conditional cmake install/export

malytomas:
One export() was not skipped when installation is disabled.
https://github.com/nih-at/libzip/issues/180

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -429,12 +429,12 @@ write_basic_package_version_file("${PROJECT_BINARY_DIR}/${PROJECT_NAME}-config-v
 configure_package_config_file("${PROJECT_NAME}-config.cmake.in" "${PROJECT_BINARY_DIR}/${PROJECT_NAME}-config.cmake"
   INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/libzip)
 
-# Add targets to the build-tree export set
-export(TARGETS zip
-  FILE "${PROJECT_BINARY_DIR}/${PROJECT_NAME}-targets.cmake")
-
-# installation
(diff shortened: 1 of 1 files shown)


0-wiz-0:
Can you explain the problem without your patch?
`make install` does not install any files for me even now.

malytomas:
Cmake would complain that there are exported targets, even if not installed, that have interface_include_directories pointing into source tree, which makes them not relocatable.
A proper solution (for me) would be to make dozens of third-party libraries separate their include directories into install and build generator expressions. This is unfeasible because some other libraries require support for much older versions of cmake.
You could, of course, do that, but a much simpler solution is to just skip the export alongside the installation.
I hope this makes sense?

Do you have any reason to keep the export if not for install?

0-wiz-0:
I have no problem with moving the code, but I thought that the export file and the install step are separate things and don't need to be coupled.

In particular, I thought I had solved the problem you refer to in libzip itself, see here:
https://github.com/nih-at/libzip/blob/master/lib/CMakeLists.txt#L179

What is missing?

malytomas:
You are right and you are awesome :D

I should not write comments from memory, I apologize for that.
Anyway, the point stands - it is actually zip's dependency:

```
CMake Error in externals/cage/externals/libzip/libzip/CMakeLists.txt:
  export called with target "zip" which requires target "zlib" that is not in
  any export set.
```

Again, I could go and try to make all the other libraries (zlib, zstd, all the crypto libs, etc.) fix build/install include directories, or make export/install optional. Fixing their include directories is not possible in older cmake versions, whereas making installation optional is generally useful.

I hope this convinces you that the patch is justifiable.

0-wiz-0:
Thanks, the problem is clearer now. I've merged it!

PR #213, opened 2020-09-26
Fix cmake path assumptions

r-burns:
Fixes downstream packaging issue https://github.com/NixOS/nixpkgs/pull/90610#issuecomment-654034293

Paths from GNUInstallDirs are not necessarily relative, in which case
they can't be simply appended to the install prefix in the .pc file.
See https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files

Also fixes hardcoded include directory.


cmake/JoinPaths.cmake
```diff
@@ -0,0 +1,23 @@
+# This module provides function for joining paths
+# known from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
+# Copyright 2020 Jan Tojnar
+# https://github.com/jtojnar/cmake-snips
+#
+# Modelled after Python’s os.path.join
+# https://docs.python.org/3.7/library/os.path.html#os.path.join
+# Windows not supported
```
0-wiz-0, line 10: That needs to be fixed if you want this to be merged.

0-wiz-0:
Hm, then perhaps we should use `CMAKE_INSTALL_FULL_<dir>` in the pkgconfig file instead.

cmake/JoinPaths.cmake
r-burns: Oh sorry, I wasn't aware of that. I'm not a Windows developer, but I've written a fair bit of CMake so I'm happy to implement a fix if I can be pointed to the relevant documentation. Is it simply a matter of changing the path separator to "\\"? Pinging @jtojnar

r-burns:
I think it's preferred to have the .pc files interpolate relative to the installation prefix when applicable - cmake-snips mentions this briefly but doesn't go into the details of why.

cmake/JoinPaths.cmake
jtojnar: [Windows absolute paths](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats) start with a drive letter. If `IS_ABSOLUTE` correctly detect them ([CMake docs](https://cmake.org/cmake/help/latest/command/if.html) are not very detailed on it), these should work, otherwise it behaves just like before.

Second, there are also root-relative paths, those would need to be properly joined by a string manipulation: leave the old drive letter but replace the rest of the path. If CMake’s `IS_ABSOLUTE` correctly detects them as relative this will work incorrectly like before. If CMake incorrectly detects root-relative paths as absolute, it will produce incorrect path. (I would expect CMake to work correctly but some Windows user would have to check or inspect the CMake source code.)

Last, there are relative paths and these should work correctly as before (unless CMake detects them as absolute too.)

Using forward slash for directory separators should be fine for most cases. It would be nice to use proper platform-specific ones but this is not a regression.

So in our use case it works for Windows as before (barring a CMake bug), even though I would not expect Windows users to use pkg-config.

Since a [merge request adding path manipulations functions](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5158) have been merged to CMake this month and the fact that it should work in our use case I do not see a point in implementing proper Windows support.

r-burns: Awesome, thanks for the detailed response.
To clarify the behavior of IS_ABSOLUTE, CMake does the right thing on windows as early as v3.0.2 - anything starting with \<driveletter>: or a leading slash will be recognized as an absolute path. (They could certainly be more clear about this in the documentation so you don't have to dig through the source code :P)
https://github.com/Kitware/CMake/blob/v3.0.2/Source/cmIfCommand.cxx#L521
https://github.com/Kitware/CMake/blob/v3.0.2/Source/kwsys/SystemTools.cxx#L3980

jtojnar: Hmm, but it looks like the root-relative paths are incorrectly detected as absolute: https://github.com/Kitware/CMake/blob/a00f1beb530c0c9fff8deab150a07b4c3839285b/Source/kwsys/SystemTools.cxx#L3616-L3619

But I would not expect root-relative paths to be used often, and pkg-config with them even less.

r-burns: Haha, well the MS link you posted refers to those as both "relative from the root of the current drive" and "An absolute path from the root of the current drive" so I would suppose they are technically both absolute _and_ relative. Either way, you are probably correct, I can't imagine such a path ever making sense from pkgconfig's point of view.

r-burns:
Ping, any more thoughts on this? If I understand correctly, this will work correctly on Windows too. Maybe I should remove or clarify the `Windows not supported` comment so it isn't misleading?

0-wiz-0:
I think the correct solution is to change:
````
SET(bindir \${exec_prefix}/${CMAKE_INSTALL_BINDIR})
SET(libdir \${exec_prefix}/${CMAKE_INSTALL_LIBDIR})
SET(includedir \${prefix}/${CMAKE_INSTALL_INCLUDEDIR})
````
to
````
SET(bindir ${CMAKE_FULL_INSTALL_BINDIR})
SET(libdir ${CMAKE_FULL_INSTALL_LIBDIR})
SET(includedir ${CMAKE_FULL_INSTALL_INCLUDEDIR})
````
Does this work for you?

If not, can you please provide the full cmake invocation?

r-burns:
Ok, I think that's a good call, the cmake is much simpler if just using the full path without interpolations is acceptable. Works for me, pushed changes.

0-wiz-0:
Thanks!

PR #216, opened 2020-10-20
[Windows] Provide meaningful error message for PATH_NOT_FOUND

cmb69:
`ENOENT` means no such file or directory, which on Windows is either
`FILE_NOT_FOUND` or `PATH_NOT_FOUND`.  I've noticed the meaningless
"Unknown error" while working on <https://bugs.php.net/80261>, where
the overlong path caused `GetFileAttributesExW()` to fail with
`PATH_NOT_FOUND`.


dillof:
Thanks, merged.

PR #225, opened 2020-12-14
Move variable declaration to support old compiler

boaz001:
Additionally I also had to add the <limits.h> include for INT_MAX and ULONG_MAX.


0-wiz-0:
Thanks, merged.

PR #227, opened 2020-12-27
Instead of listing all tests, use glob to capture all

### LINKED ISSUE #226: set_file_mtime.test is not executed
Test set_file_mtime.test is not executed because it is missing in the regress/CMakeLists.txt

boaz001:
Closes #226

This solution uses a file pattern: `*.test` to find all tests in the regress directory instead of manually listing them.


0-wiz-0:
Thanks, I used your change as the base for my commit!

PR #229, opened 2021-01-06
LibreSSL has used the OpenSSL 1.1 API since 2.7

vishwin:
LibreSSL 2.7.0 was released 21 March 2018, just after #35 landed. While continuing to force use of the OpenSSL 1.0 API didn't result in any build or runtime errors, since 2.7 LibreSSL has implemented the OpenSSL 1.1 API.


0-wiz-0:
Thanks, merged (untested, since I don't have a libressl installation).

PR #240, opened 2021-05-28
Don't define snprintf macro under MSVC

phandasm:
Newer versions of Visual Studio explicitly forbid defining snprintf as a macro and #error out.
This PR disables the macro if snprintf proper is present.


0-wiz-0:
Thank you!

PR #244, opened 2021-06-18
CMakeLists.txt: add missing regex for zstd

rezso:
Currently cmake provides an pc file with:
Libs.private:  -lbz2 -llzma -lZstd::Zstd -lz


0-wiz-0:
Thank you, merged!

PR #248, opened 2021-06-24
Fix paths in pkgconfig file

mhei:
Commit 9e2e7178 tried to address an issue with wrong path assumptions
about relative vs. absolute directories and referenced a good explanation
how to do things correctly.

However, the commit itself does not follow the recommendations and this
leads to hard-coded full paths in the pkgconfig file - which is wrong
in my eyes and also breaks the build of dependend packages on e.g.
OpenWrt - I'd also call it regression since previous libzip releases
worked well.

This commit tries to fixup things by following the recommendation
of the already referenced page:
https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files

Notes:
1. The included cmake snippet is licensed under MIT or CC0-1.0 which
is compatible in my eyes for inclusion.
2. The included cmake snippet is not Windows compatible, and I do not have
a Windows platform to test on hand.

Signed-off-by: Michael Heimpold <mhei@heimpold.de>


0-wiz-0:
It was not immediately clear what you are requesting.
My understanding is that you want to override pkg-config's ${prefix} after installation time, and that did not work because we used absolute paths for includedir etc.
The submitted patch does not address this though.

I've committed a change that tries to make the three variables bindir/libdir/includedir relative paths.
(if any of these is not below prefix, this will now generate broken pkg-config files, but that is just not supported).

mhei:
Your understanding was correct. My problem was, that I did not fully understand, what the offending commit 9e2e717  tried to achieve.

> The submitted patch does not address this though.

Sure, it does - however, your solution seems to work, too - so I'm happy.
Thanks for your efforts.

PR #251, opened 2021-07-18
CMake: fix install destination of executables for iOS/tvOS/watchOS

SpaceIm:
When tools are enabled, CMake configuration fails if host is iOS, tvOS or watchOS. For these hosts, executables are BUNDLE, not RUNTIME, and they need a destination. Since this install() command only refers to executables, we can globally set DESTINATION for all types, and it's done.

see https://cmake.org/cmake/help/latest/policy/CMP0006.html


dillof:
Do you have any use for these executables on iOS, tvOS, or watchOS? If not, disabling installation of the executables on those platforms seems like a better approach to me.

SpaceIm:
> Do you have any use for these executables on iOS, tvOS, or watchOS?

No

> If not, disabling installation of the executables on those platforms seems like a better approach to me.

Sure, but this simple fix avoids consumers to hack default options values, and it has no drawbacks.

dillof:
I'll look into setting the default to not installing executables on those platforms.

I'm not so sure about the drawbacks. Usually when we change the way things are installed, we immediately hear from some other platform where it no longer works. I'll look into this as well.

SpaceIm:
https://cmake.org/cmake/help/latest/command/install.html

> For each of these arguments given, the arguments following them only apply to the target or file type specified in the argument. If none is given, the installation properties apply to all target types. If only one is given then only targets of that type will be installed (which can be used to install just a DLL or just an import library.)

dillof:
Sounds harmless enough, here's hoping there is no obscure edge case this is triggering somewhere…

PR #257, opened 2021-09-06
fix linking errors without SSL

neheb:
These function are not built without SSL.


0-wiz-0:
Can you please provide more details about the error you see?
I wasn't aware that linkers complain about function that are only listed in header files. Which linker is that?
Do you need special flags to enable that behaviour?
Thanks.

neheb:
This linker is ccache 4.4 :).

It's actually an upstream bug: https://github.com/ccache/ccache/issues/926

However, the fix is simple enough.

0-wiz-0:
That bug is closed as irreproducible...

neheb:
Oh it's very easy to reproduce.

0-wiz-0:
You don't have to convince me about that, but the ccache people :)

dillof:
This patch is incomplete, as there are various other functions and variables that are only defined if the corresponding libraries were found by cmake.

I don't feel completing and maintaining this information in the header file is worth the effort, as it should not cause any problems and usually doesn't.

PR #272, opened 2021-12-06
zip_mkstempm: Add flag to allow using a custom suffix 

easyteacher:
There could be such a scenario for new users:

1. Compress many files or large files.
2. A temporary file is created.
3. The user doesn't know it's a temporary file, and continuously tries to open it.

The situation has happened in a video: https://youtu.be/TtsglXhbxno?t=720

We should give the temporary file a descriptive suffix, and ".part" seems to be the most reasonable and familiar suffix for most users, because the suffix is also used by many browsers to indicate a file is still being downloaded.

To make the solution clean, `_zip_mkstempm` is also adapted to three use cases:

1. Default mode      (mkstemp)
2. Fill in name only (mktemp)
3. Fill in name only, but with a custom suffix

When calling `_zip_stdio_op_create_temp_output`, the third flag is used so a custom suffix is made possible.

### DIFF
--- a/lib/zip_mkstempm.c
+++ b/lib/zip_mkstempm.c
@@ -41,16 +41,34 @@
 /*
  * create temporary file with same permissions as previous one;
  * or default permissions if there is no previous file
+ *
+ * @param path File path
+ * @param mode File permission
+ * @param flag 0 - Default mode      (mkstemp)
+ *             1 - Fill in name only (mktemp)
+ *             2 - Fill in name only, but with a custom suffix
  */
 int
-_zip_mkstempm(char *path, int mode) {
+_zip_mkstempm(char *path, int mode, unsigned int flag) {
+    if (flag == 0) {
+        return mkstemp(path);
+    } else if (flag == 1) {
+        return mktemp(path) != NULL ? 1 : -1;
+    }
+
     int fd;
     char *start, *end, *xs;
 
-    int xcnt = 0;
+    int xcnt = 0, pos = 0;
 
     end = path + strlen(path);
     start = end - 1;
+
+    while (start >= path && *start != 'X') {
+        start--;
+        pos++;
+    }
+
     while (start >= path && *start == 'X') {
         xcnt++;
         start--;
@@ -68,7 +86,7 @@ _zip_mkstempm(char *path, int mode) {
 
         xs = start;
 
-        while (xs < end) {
+        while (xs < end - pos) {
             char digit = value % 36;
             if (digit < 10) {
                 *(xs++) = digit + '0';
--- a/lib/zip_source_file_stdio_named.c
+++ b/lib/zip_source_file_stdio_named.c
@@ -129,7 +129,7 @@ _zip_stdio_op_create_temp_output(zip_source_file_context_t *ctx) {
     FILE *tfp;
     struct stat st;
 
-    if ((temp = (char *)malloc(strlen(ctx->fname) + 8)) == NULL) {
+    if ((temp = (char *)malloc(strlen(ctx->fname) + 13)) == NULL) {
         zip_error_set(&ctx->error, ZIP_ER_MEMORY, 0);
         return -1;
     }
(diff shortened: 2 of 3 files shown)


dillof:
While I support the goal, this patch has issues:

1. It does not work when cloning is supported (currently on macOS with APFS and Linux with BTRFS). Both of these code paths use standard library functions to create the temporary file name, which expects Xs at the end.
2. _zip_mkstempm() no longer behaves like mkstemp() but assumes that there are exactly 5 characters after the Xs to be replaced. A better approach would be to look for the last string of Xs and replace those.

easyteacher:
>It does not work when cloning is supported (currently on macOS with APFS and Linux with BTRFS). Both of these code paths use standard library functions to create the temporary file name, which expects Xs at the end.

Thank you for the feedback. Do you mean I should not add ".part" to `_zip_stdio_op_create_temp_output_cloning` or it will cause issues?

>_zip_mkstempm() no longer behaves like mkstemp() but assumes that there are exactly 5 characters after the Xs to be replaced. A better approach would be to look for the last string of Xs and replace those.

Will do.

dillof:
> > It does not work when cloning is supported (currently on macOS with APFS and Linux with BTRFS). Both of these code paths use standard library functions to create the temporary file name, which expects Xs at the end.
> 
> Thank you for the feedback. Do you mean I should not add ".part" to `_zip_stdio_op_create_temp_output_cloning` or it will cause issues?

libzip should behave the same in both cases, otherwise it would be even more confusing to the user. But that means you can no longer use the standard library functions. Probably the cleanest solution would be to adapt _zip_mkstempm to all use cases:

1. create temporary file with given mode
2. create temporary file with default mode
3. don't create temporary file, fill in name only

and then use only that function and none mktemp / mkstemp.
> 
> > _zip_mkstempm() no longer behaves like mkstemp() but assumes that there are exactly 5 characters after the Xs to be replaced. A better approach would be to look for the last string of Xs and replace those.
> 
> Will do.

Thanks.

easyteacher:
Done. Please review the change.

dillof:
I've taken this opportunity to further clean up the relevant code paths, and implemented the change on the branch part-temp-name. 

However, I can't test the change on Linux with BTRFS, which is required for cloning support. If someone can test that, please let me know (simply build the branch part-temp-name and run `make check`). Once it's confirmed to work there, we'll merge it into the main branch.

shenlebantongying:
@dillof I tested on btrfs from linux kernal version 5.15.7, and it says all passed
image

dillof:
@shenlebantongying Great, thank you. I've merged the change into the main branch.

easyteacher:
So now we can also use .part suffix when btrfs cloning is enabled, is that right?

dillof:
Yes, now the temporary file name for `directory/file.zip` is `directory/.file.zip.XXXXXX.part` (where XXXXXX is replaced by random characters). It's the same for copying, macOS cloning, and Linux cloning.

easyteacher:
How can I see if btrfs cloning is enabled? I also tested my patch on btrfs, but it seems .part suffix is also added.

dillof:
If you run `make check` on a system without btrfs, you'll see the lines:

```
The following tests did not run:
         24 - clone-fs-add.test (Skipped)
         25 - clone-fs-delete.test (Skipped)
         26 - clone-fs-replace.test (Skipped)
```

If those tests result in `Passed` rather than `Skipped`, you have btrfs.

easyteacher:
I also have `100% tests passed, 0 tests failed out of 145`. But I would like to know if I compress a new file, will btrfs cloning work?

easyteacher:
OK I understand now. Only when adding files to an existing zip archive will btrfs cloning work.

https://bugs.kde.org/show_bug.cgi?id=446496 is closed now.

dillof:
Yes, basically. We don't modify the original file, we write a completely new one so that there is a valid version of the zip archive at all times. If you don't modify the first N files in the archive, these have to be copied unchanged to the new file. If the file system allows cloning (basically a hardlink with copy-on-write), we clone the unchanged portion and only write the new files, which is obviously much more efficient. Unchanged files after a changed file will most likely change position inside the archive, so they are always copied.

easyteacher:
Thank you very much for maintaining such an important library, and thanks for the explaining!

PR #273, opened 2021-12-11
Don't hide the temp file

easyteacher:
According to [the discussion](https://invent.kde.org/teams/usability/issue-board/-/issues/2#note_355542), hiding the temp file can be problematic because:

1. If the compression operation fails for some reason (disk full,
   segfault, etc) the user won't be able to see the file directly
   and is unlikely to delete it.
2. Lack of instantaneous feedback that something is happening.

make check on btrfs: 100% tests passed, 0 tests failed out of 145

### DIFF
--- a/lib/zip_source_file_stdio_named.c
+++ b/lib/zip_source_file_stdio_named.c
@@ -301,7 +301,7 @@ static int create_temp_file(zip_source_file_context_t *ctx, bool create_file) {
         mode = -1;
     }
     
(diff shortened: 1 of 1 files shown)


dillof:
While I only partially agree with your arguments (progress feedback should be provided by the application using libzip; in case of disk full, the temp file is deleted by libzip), it seems unusual for Unix programs to hide temp files, so we are now following that convention again.

PR #285, opened 2022-02-21
Fix unchange after deletion of the original file

krnowak:
Unchange didn't revert setting current_index to -1 in names hash, so
the name could not be located any more.


dillof:
Thanks, we fixed the underlying bug in a slightly different way.

krnowak:
Cool, thanks!

PR #286, opened 2022-02-26
Allow changed reads

krnowak:
Hi,

I'd like to show you some PoC code that allows reading contents of the changed files in the archive (that is - doing `zip_fopen` without the `ZIP_FL_UNCHANGED` flag). This is implemented as an opt-in feature of `zip_source` - it needs to say that it supports the `ZIP_SOURCE_ALLOW_CHANGED_READ` command, and that command should return 1 if it's actually allowed. In such case, `zip_fread` will use the source to read data from it, instead of returning `ZIP_ER_CHANGED`. The feature should be backwards compatible, as no builtin non-layered sources support this command, and the builtin layered sources (encryption, window and crc) are deferring to the underlying source.

My usecase was basically to have the zip archive to serve as a backend (next to the "directory and files" backend) for a stream source, which can provide input or output streams. This change would allow me to open an output stream, write some data to it, and then close it; later I could open the same stream, but as an input and read from it.

Would such a feature be considered for inclusion to the project?


dillof:
Thank you. The functionality sounds good, but the implementation will need some refinement.

We'll discuss it internally and get back to you, probably some time next week.

krnowak:
> Thank you. The functionality sounds good, but the implementation will need some refinement.

Cool, good to know, thanks. About implementation - surely it will need refinement. Just for the record - I tried extending `_zip_source_zip_new` but it was rather a frustrating exercise, as changing/adapting something in this function always seemed to break that or another test case. That's why I went with the duplication of code - I assumed it should be just fine for the PoC code.

> 
> We'll discuss it internally and get back to you, probably some time next week.

Alright, I'll fix the errors from CI in the meantime.

dillof:
The feature you add to sources has broader application than reading changed data, so a more general name would better describe it. We've come up with `ZIP_SOURCE_SUPPORTS_REOPEN`.

Also, by moving support for it into the infrastructure, we can minimise diffs for each source:

```diff --git a/lib/zip_source_layered.c b/lib/zip_source_layered.c
index f6db2a71..d59ab828 100644
--- a/lib/zip_source_layered.c
+++ b/lib/zip_source_layered.c
@@ -62,6 +62,9 @@ zip_source_layered_create(zip_source_t *src, zip_source_layered_callback cb, voi
     if (zs->supports < 0) {
         zs->supports = ZIP_SOURCE_SUPPORTS_READABLE;
     }
+    else if ((zip_source_supports(src) & ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN)) == 0) {
+        zs->supports &= ~ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN);
+    }
 
     return zs;
 }
diff --git a/lib/zip_source_supports.c b/lib/zip_source_supports.c
index 66a5303c..8ccc41bc 100644
--- a/lib/zip_source_supports.c
+++ b/lib/zip_source_supports.c
@@ -63,3 +63,8 @@ zip_source_make_command_bitmap(zip_source_cmd_t cmd0, ...) {
 
     return bitmap;
 }
+
+
+bool zip_supports_multiple_open(zip_source_t *src) {
+    return (zip_source_supports(src) & ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN)) != 0;
+}
```

Also, we don't need to make another call that only ever returns 1. If a source says it supports reopen, then no further checks are needed.

With that, all that's needed in each source is to add `ZIP_SOURCE_SUPPORTS_REOPEN` to the bitmap of supported commands in `ZIP_SOURCE_SUPPORTS`.

Please do not duplicate the main part of `zip_source_zip_new`. Make a function you can call with a `zip_source_t` and a `zip_stat_t` describing the format of the existing data source and `flags` that describe the desired data format; this function should figure out which layers need to be put on top to make that happen (this is the part you duplicated); then call that function from both cases. 

Also, please don't add a special case to `fread.c`. For most tests we use `ziptool_regress`, which should support all of the functions you need. (The `cat` sub-command already reads changed data.) Please use that in your tests.

On a stylistic note, please remove spaces before opening parentheses in function calls (`function(arguments)` rather than `function (arguements)`).

krnowak:
> The feature you add to sources has broader application than reading changed data, so a more general name would better describe it. We've come up with `ZIP_SOURCE_SUPPORTS_REOPEN`.

This is certainly a better name. Will use it, thanks.

> 
> Also, by moving support for it into the infrastructure, we can minimise diffs for each source:
> 
> ```diff
> index f6db2a71..d59ab828 100644
> --- a/lib/zip_source_layered.c
> +++ b/lib/zip_source_layered.c
> @@ -62,6 +62,9 @@ zip_source_layered_create(zip_source_t *src, zip_source_layered_callback cb, voi
>      if (zs->supports < 0) {
>          zs->supports = ZIP_SOURCE_SUPPORTS_READABLE;
>      }
> +    else if ((zip_source_supports(src) & ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN)) == 0) {
> +        zs->supports &= ~ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN);
> +    }
>  
>      return zs;
>  }
> diff --git a/lib/zip_source_supports.c b/lib/zip_source_supports.c
> index 66a5303c..8ccc41bc 100644
> --- a/lib/zip_source_supports.c
> +++ b/lib/zip_source_supports.c
> @@ -63,3 +63,8 @@ zip_source_make_command_bitmap(zip_source_cmd_t cmd0, ...) {
>  
>      return bitmap;
>  }
> +
> +
> +bool zip_supports_multiple_open(zip_source_t *src) {
> +    return (zip_source_supports(src) & ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN)) != 0;
> +}
> ```
> 
> Also, we don't need to make another call that only ever returns 1. If a source says it supports reopen, then no further checks are needed.
> 
> With that, all that's needed in each source is to add `ZIP_SOURCE_SUPPORTS_REOPEN` to the bitmap of supported commands in `ZIP_SOURCE_SUPPORTS`.

This is a nice simplification, thanks. The only thing I thought is that the source could decide during `ZIP_SOURCE_SUPPORTS_REOPEN` command execution whether the reopening is possible at this moment (because we are still appending to the internal buffer or something), but I suppose I can return an error from `ZIP_SOURCE_OPEN` or `ZIP_SOURCE_STAT`.

> 
> Please do not duplicate the main part of `zip_source_zip_new`. Make a function you can call with a `zip_source_t` and a `zip_stat_t` describing the format of the existing data source and `flags` that describe the desired data format; this function should figure out which layers need to be put on top to make that happen (this is the part you duplicated); then call that function from both cases.
> 
> Also, please don't add a special case to `fread.c`. For most tests we use `ziptool_regress`, which should support all of the functions you need. (The `cat` sub-command already reads changed data.) Please use that in your tests.
> 

Will try it out.

> On a stylistic note, please remove spaces before opening parentheses in function calls (`function(arguments)` rather than `function (arguements)`).

Oops, sorry about that. My personal style got in. Will fix it.

dillof:
> > Also, we don't need to make another call that only ever returns 1. If a source says it supports reopen, then no further checks are needed.
> > With that, all that's needed in each source is to add `ZIP_SOURCE_SUPPORTS_REOPEN` to the bitmap of supported commands in `ZIP_SOURCE_SUPPORTS`.
> 
> This is a nice simplification, thanks. The only thing I thought is that the source could decide during `ZIP_SOURCE_SUPPORTS_REOPEN` command execution whether the reopening is possible at this moment (because we are still appending to the internal buffer or something), but I suppose I can return an error from `ZIP_SOURCE_OPEN` or `ZIP_SOURCE_STAT`.
> 

We need a promise from the source that it will allow a second open, otherwise `zip_close()` will fail. So the source has to decide up front if it can support that, otherwise it should simply say it doesn't. Having `zip_close()` succeed is much more important than allowing reading changed data.

krnowak:
> Also, by moving support for it into the infrastructure, we can minimise diffs for each source:
> 
> ```diff
> index f6db2a71..d59ab828 100644
> --- a/lib/zip_source_layered.c
> +++ b/lib/zip_source_layered.c
> @@ -62,6 +62,9 @@ zip_source_layered_create(zip_source_t *src, zip_source_layered_callback cb, voi
>      if (zs->supports < 0) {
>          zs->supports = ZIP_SOURCE_SUPPORTS_READABLE;
>      }
> +    else if ((zip_source_supports(src) & ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN)) == 0) {
> +        zs->supports &= ~ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN);
> +    }
>  
>      return zs;
>  }
> ```

I also added the case `if (source supports reopen) zs->supports |= SUPPORTS_REOPEN` here, so the layered sources support reopening if the underlying source also supports it.

> 
> Please do not duplicate the main part of `zip_source_zip_new`. Make a function you can call with a `zip_source_t` and a `zip_stat_t` describing the format of the existing data source and `flags` that describe the desired data format; this function should figure out which layers need to be put on top to make that happen (this is the part you duplicated); then call that function from both cases.

In the end I extended the `zip_source_zip_new` function, because the new function would basically be taking all the parameters of the `zip_source_zip_new` to determine the layers (`srcza`, `srcidx`, `start`, `len` would be needed for windowed layer, `start` and `len` for determining the partial read, `password` for decryption layer, so meh). But the additions were not that messy in the end (I think).

> 
> Also, please don't add a special case to `fread.c`. For most tests we use `ziptool_regress`, which should support all of the functions you need. (The `cat` sub-command already reads changed data.) Please use that in your tests.
> 

I replaced that with ziptool, but needed to add the special "{add,replace}_reopenable" command to ziptool-regress, as I haven't added support for reopening to any of the builtin sources yet. I'm not sure how to do it yet (either add `ZIP_SOURCE_SUPPORTS_REOPEN` to the supports set, or add `zip_file_{add,replace}_reopenable` (and similar for buffer source) or add ZIP_FL_REOPENABLE to `zip_flags_t`), so I decided to leave it out and add a simple reopenable read-only buffer source to ziptool regress.

TODO:

- update docs
- add a test that adds a reopenable file to a zip archive, and then adds a part of the reopenable file to another zip archive

lib/zip_source_layered.c
```diff
@@ -62,6 +62,12 @@ zip_source_layered_create(zip_source_t *src, zip_source_layered_callback cb, voi
     if (zs->supports < 0) {
         zs->supports = ZIP_SOURCE_SUPPORTS_READABLE;
     }
+    if (zip_source_supports_reopen(src)) {
+        zs->supports |= ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN);
```
dillof, line 66: Just because the lower layer supports reopening doesn't mean the top layer does, too. A source must explicitly advertise support for reopening. Please remove this line.

lib/zip_source_zip_new.c
```diff
@@ -43,20 +43,43 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
...
+    if (!reopening) {
+        stat_flags |= ZIP_FL_UNCHANGED;
     }
 
-    if (zip_stat_index(srcza, srcidx, flags | ZIP_FL_UNCHANGED, &st) < 0) {
-        zip_error_set(error, ZIP_ER_INTERNAL, 0);
+    if (zip_stat_index(srcza, srcidx, stat_flags, &st) < 0) {
+        /* if reopening, the error could come from 3rd party code, so it's not internal. */
+        int ze = (reopening ? ZIP_ER_INVAL : ZIP_ER_INTERNAL);
+        zip_error_set(error, ze, 0);
```
dillof, line 82: It's probably best to just return the error from the source unchanged. (Also, ZIP_ER_INVAL is for invalid arguments, so doesn't fit here.)

lib/zip_source_zip_new.c
```diff
@@ -69,6 +92,13 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
         return NULL;
     }
 
+    /* stat may come from third party code, we need to validate it */
+    if ((st.valid & ZIP_STAT_SIZE) == 0) {
+        int ze = (reopening ? ZIP_ER_INVAL : ZIP_ER_INTERNAL);
+        zip_error_set(error, ze, 0);
+        return NULL;
+    }
```
dillof, line 100: It's not an error if we don't know the size, we'll just have to deal with it. Some checks won't be possible later on, but that's fine.

lib/zip_source_zip_new.c
```diff
@@ -43,20 +43,43 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
...
-        return NULL;
+    if ((flags & ZIP_FL_UNCHANGED) == 0) {
+        zip_entry_t *entry = srcza->entry + srcidx;
+        if (ZIP_ENTRY_DATA_CHANGED(entry)) {
+            if (!zip_source_supports_reopen(entry->source)) {
+                zip_error_set(error, ZIP_ER_CHANGED, 0);
+                return NULL;
+            }
+
+            reopening = true;
```
dillof, line 64: I don't like the name `reopening`, maybe `changed_data`?

lib/zip_source_zip_new.c
```diff
@@ -80,10 +110,12 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
...
     partial_data = len < st.size;
-    needs_decrypt = ((flags & ZIP_FL_ENCRYPTED) == 0) && (st.encryption_method != ZIP_EM_NONE);
-    needs_decompress = ((flags & ZIP_FL_COMPRESSED) == 0) && (st.comp_method != ZIP_CM_STORE);
+    encrypted = (st.valid & ZIP_STAT_ENCRYPTION_METHOD) && (st.encryption_method != ZIP_EM_NONE);
+    needs_decrypt = ((flags & ZIP_FL_ENCRYPTED) == 0) && encrypted;
+    compressed = (st.valid & ZIP_STAT_COMP_METHOD) && (st.comp_method != ZIP_CM_STORE);
+    needs_decompress = ((flags & ZIP_FL_COMPRESSED) == 0) && compressed;
     /* when reading the whole file, check for CRC errors */
-    needs_crc = ((flags & ZIP_FL_COMPRESSED) == 0 || st.comp_method == ZIP_CM_STORE) && !partial_data;
+    needs_crc = ((flags & ZIP_FL_COMPRESSED) == 0 || !compressed) && !partial_data;
```
dillof, line 118: When reopening, `needs_crc` should be false, since we don't have a CRC to compare it against. (`needs_crc` is used to compare the CRC from the actual data with the one recorded in the zip archive.)

lib/zip_source_zip_new.c
```diff
@@ -100,29 +132,50 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
     }
     _zip_file_attributes_from_dirent(&attributes, de);
 
-    if (st.comp_size == 0) {
+    if (st.size == 0) {
```
dillof, line 135: This is wrong. You have to check if the data we want to read is 0 bytes long. If we're reading compressed data, that's st.comp_size, otherwise is st.size. If that isn't known, assume it's not 0.

lib/zip_source_zip_new.c
```diff
@@ -100,29 +132,50 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
...
+            st2.mtime = st.mtime;
+            st2.valid |= ZIP_STAT_MTIME;
+        }
 
-        if ((src = _zip_source_window_new(srcza->src, start, (zip_int64_t)len, &st2, &attributes, srcza, srcidx, error)) == NULL) {
+        if ((src = _zip_source_window_new(src, start, (zip_int64_t)len, &st2, &attributes, source_archive, source_index, error)) == NULL) {
             return NULL;
         }
     }
     else {
```
dillof, line 163: This else branch only applies if not reopening: It restricts src to the file data, while srcza->src is the complete archive. When reopening, src is already just the file data.

lib/zip_source_zip_new.c
```diff
@@ -43,20 +43,43 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
     zip_stat_t st;
     zip_file_attributes_t attributes;
     zip_dirent_t *de;
-    bool partial_data, needs_crc, needs_decrypt, needs_decompress;
+    bool partial_data, needs_crc, encrypted, needs_decrypt, compressed, needs_decompress, reopening;
+    zip_flags_t stat_flags;
+    zip_t *source_archive;
+    zip_uint64_t source_index;
```
dillof, line 49: We need to make sure the returned source is invalidated when `srcza` is closed. I'm not sure what the best way to do that is, so please just add a TODO comment so we won't forget.

dillof (changes requested):

dillof:
> > Also, by moving support for it into the infrastructure, we can minimise diffs for each source:
> > ```diff
> > index f6db2a71..d59ab828 100644
> > --- a/lib/zip_source_layered.c
> > +++ b/lib/zip_source_layered.c
> > @@ -62,6 +62,9 @@ zip_source_layered_create(zip_source_t *src, zip_source_layered_callback cb, voi
> >      if (zs->supports < 0) {
> >          zs->supports = ZIP_SOURCE_SUPPORTS_READABLE;
> >      }
> > +    else if ((zip_source_supports(src) & ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN)) == 0) {
> > +        zs->supports &= ~ZIP_SOURCE_MAKE_COMMAND_BITMASK(ZIP_SOURCE_SUPPORTS_REOPEN);
> > +    }
> >  
> >      return zs;
> >  }
> > ```
> 
> I also added the case `if (source supports reopen) zs->supports |= SUPPORTS_REOPEN` here, so the layered sources support reopening if the underlying source also supports it.

That's not a safe assumption, see my review comment.

> 
> > Please do not duplicate the main part of `zip_source_zip_new`. Make a function you can call with a `zip_source_t` and a `zip_stat_t` describing the format of the existing data source and `flags` that describe the desired data format; this function should figure out which layers need to be put on top to make that happen (this is the part you duplicated); then call that function from both cases.
> 
> In the end I extended the `zip_source_zip_new` function, because the new function would basically be taking all the parameters of the `zip_source_zip_new` to determine the layers (`srcza`, `srcidx`, `start`, `len` would be needed for windowed layer, `start` and `len` for determining the partial read, `password` for decryption layer, so meh). But the additions were not that messy in the end (I think).

Yes, that approach works fine.

> 
> > Also, please don't add a special case to `fread.c`. For most tests we use `ziptool_regress`, which should support all of the functions you need. (The `cat` sub-command already reads changed data.) Please use that in your tests.
> 
> I replaced that with ziptool, but needed to add the special "{add,replace}_reopenable" command to ziptool-regress, as I haven't added support for reopening to any of the builtin sources yet. I'm not sure how to do it yet (either add `ZIP_SOURCE_SUPPORTS_REOPEN` to the supports set, or add `zip_file_{add,replace}_reopenable` (and similar for buffer source) or add ZIP_FL_REOPENABLE to `zip_flags_t`), so I decided to leave it out and add a simple reopenable read-only buffer source to ziptool regress.

Just advertise support for reopening to sources that can handle it. That should be most of the built in sources.

> 
> TODO:
> 
> * update docs
> * add a test that adds a reopenable file to a zip archive, and then adds a part of the reopenable file to another zip archive

That's a more complicated test than necessary. For now, just add a file and read it with zip_tool cat.

lib/zip_source_zip_new.c
krnowak: So the CRC source gets the checksum from underlying source by calling `zip_source_stat`, so I thought that if the custom source actually returns CRC in its stat object, why blocking the use of it?

lib/zip_source_zip_new.c
krnowak: Maybe by calling `_zip_source_set_source_archive` on each layer? But I'll add a TODO.

lib/zip_source_zip_new.c
krnowak: I'll make if `else if (!changed_data) {` then and add a comment here. Essentially this branch will be only executed for `srcza->src`, so should I revert the changes I made for this branch? (Can we assume that `st` is going to have a valid `comp_size` member set?)

krnowak:
> > I replaced that with ziptool, but needed to add the special "{add,replace}_reopenable" command to ziptool-regress, as I haven't added support for reopening to any of the builtin sources yet. I'm not sure how to do it yet (either add `ZIP_SOURCE_SUPPORTS_REOPEN` to the supports set, or add `zip_file_{add,replace}_reopenable` (and similar for buffer source) or add ZIP_FL_REOPENABLE to `zip_flags_t`), so I decided to leave it out and add a simple reopenable read-only buffer source to ziptool regress.
> 
> Just advertise support for reopening to sources that can handle it. That should be most of the built in sources.
> 

Ok, will do - at first I didn't want to change the built in sources (I dunno, someone relies on getting `ZIP_ER_CHANGED` from `zip_fopen` after adding a buffer source there?), but will update the support bitmasks.

> > TODO:
> > 
> > * update docs
> > * add a test that adds a reopenable file to a zip archive, and then adds a part of the reopenable file to another zip archive
> 
> That's a more complicated test than necessary. For now, just add a file and read it with zip_tool cat.

Yeah, that's already done I'd say. With this TODO item, I wanted to test the code path where we do a partial read of the changed data and this seems to be the only way.

krnowak:
> > > * add a test that adds a reopenable file to a zip archive, and then adds a part of the reopenable file to another zip archive
> > 
> > 
> > That's a more complicated test than necessary. For now, just add a file and read it with zip_tool cat.
> 
> Yeah, that's already done I'd say. With this TODO item, I wanted to test the code path where we do a partial read of the changed data and this seems to be the only way.

Just realized that I could do it with just one archive.

lib/zip_source_zip_new.c
dillof: Yes, that makes sense. In that case add a `&& (st.valid & ZIP_STAT_CRC)` to `needs_crc` so it only gets added if we have a CRC to compare against.

lib/zip_source_zip_new.c
dillof: Yes, please revert your changes here. You can assume that `comp_size` is valid here, since the stat information is taken from the archive.

dillof:
> > > I replaced that with ziptool, but needed to add the special "{add,replace}_reopenable" command to ziptool-regress, as I haven't added support for reopening to any of the builtin sources yet. I'm not sure how to do it yet (either add `ZIP_SOURCE_SUPPORTS_REOPEN` to the supports set, or add `zip_file_{add,replace}_reopenable` (and similar for buffer source) or add ZIP_FL_REOPENABLE to `zip_flags_t`), so I decided to leave it out and add a simple reopenable read-only buffer source to ziptool regress.
> > 
> > 
> > Just advertise support for reopening to sources that can handle it. That should be most of the built in sources.
> 
> Ok, will do - at first I didn't want to change the built in sources (I dunno, someone relies on getting `ZIP_ER_CHANGED` from `zip_fopen` after adding a buffer source there?), but will update the support bitmasks.
> 
> > > TODO:
> > > 
> > > * update docs
> > > * add a test that adds a reopenable file to a zip archive, and then adds a part of the reopenable file to another zip archive
> > 
> > 
> > That's a more complicated test than necessary. For now, just add a file and read it with zip_tool cat.
> 
> Yeah, that's already done I'd say. With this TODO item, I wanted to test the code path where we do a partial read of the changed data and this seems to be the only way.

To test partial reads, we should add a command to `zip_tool` that outputs part of a file, like `cat_partial <index> <start> <length>`. We try to avoid writing programs for individual test cases.

lib/zip_source_layered.c
krnowak: Removed.

krnowak:
Updated, still need to write some docs.

> > > > * add a test that adds a reopenable file to a zip archive, and then adds a part of the reopenable file to another zip archive
> > > 
> > > 
> > > That's a more complicated test than necessary. For now, just add a file and read it with zip_tool cat.
> > 
> > 
> > Yeah, that's already done I'd say. With this TODO item, I wanted to test the code path where we do a partial read of the changed data and this seems to be the only way.
> 
> To test partial reads, we should add a command to `zip_tool` that outputs part of a file, like `cat_partial <index> <start> <length>`. We try to avoid writing programs for individual test cases.

So my idea was to have more or less something like this:

```
./ziptool testbuffer.zip \
    cat 0 \
    replace_file_contents 0 "Overwritten\n" \
    cat 0 \
    add_from_zip part.txt testbuffer.zip 0 4 5 \
    cat 1
```

But that will open testbuffer.zip again, where the modification from `replace_file_contents` does not exist. Meh.

`cat_partial <index> <start> <length>` currently can be implemented in term of fopen, fread to throwaway buffer, then fread + fwrite to stdout, then fclose. Which is not ideal, because fopen opens the source with `start` and `len` set to zeroes, and I wanted to test the partial reads paths.

krnowak:
> `cat_partial <index> <start> <length>` currently can be implemented in term of fopen, fread to throwaway buffer, then fread + fwrite to stdout, then fclose. Which is not ideal, because fopen opens the source with `start` and `len` set to zeroes, and I wanted to test the partial reads paths.

Or by `zip_source_zip` and using `zip_source_{open,read,close,free}`, maybe with some seeking. I'll have a look. Didn't notice before that zip_source_* is actually a part of API.

krnowak:
Implemented `cat_partial` in ziptool and added two more tests.

krnowak:
For now I can say that I have no idea why the tests fail on appveyor. At first I thought it's because of enabling reopen in `zip_source_file_common.c`, but removing the support didn't change a thing. I only have a feeling that in appveyor I'm getting no stderr output from the failed tests.

I also did a small stab at documentation, but I'll admit that the mdoc format is pretty much alien to me.

krnowak:
Updated, the issue with windows failures and hopefully with the fuzzer is resolved.

dillof:
Sorry, I've been busy lately, but I'll have a look at it next week.

Thank you for your continued work on this issue.

lib/zip_source_zip_new.c
```diff
@@ -126,6 +188,10 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
             return NULL;
         }
     }
+    else {
+        src = srcza->entry[srcidx].source;
+        zip_source_keep(src);
```
krnowak, line 193: Used the branch a bit more and there was one issue - having two `zip_file_t` to the same reopened source messes up reads - reading from one `zip_file_t` affects reading from the another. It normally works for unchanged sources, because window source does seeks before calling read on the underlying source. I think here we will need to wrap the reopened source into window source here too.

dillof (changes requested):

src/ziptool.c
```diff
@@ -207,6 +207,60 @@ cat(char *argv[]) {
     return 0;
 }
 
+static int
+cat_partial(char *argv[]) {
```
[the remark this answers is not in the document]
krnowak: The difference is that `cat` uses `zip_file_t`, while `cat_partial` is implemented in terms of `zip_source_t`. I suppose I can assume that reducing code coverage of `zip_file_t` in testing is ok here.

lib/zip_fclose.c
```diff
@@ -41,8 +41,10 @@ ZIP_EXTERN int
 zip_fclose(zip_file_t *zf) {
     int ret;
 
-    if (zf->src)
+    if (zf->src) {
+        zip_source_close(zf->src);
```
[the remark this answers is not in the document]
krnowak: Ok, I can revert this change. I'll only point you to the commit message in https://github.com/nih-at/libzip/pull/286/commits/4a97e13e0b30e32ef4b61a67d3f21b9046a40cdf. Now this probably won't be a problem any more, because we always wrap the bottom-most source in some other sources, which happen to have ref count of 1, so they'll be closed on free.

I'd still say that `zip_source_free` is a misnomer (should be rather something like `zip_source_unref`), because it's freeing things only if ref count drops to 0. So if refcount is still higher, no free will happen and no closing either.

lib/zip_source_zip_new.c
```diff
@@ -39,24 +39,42 @@
...
+        }
+    }
+
+    stat_flags = flags;
+    if (!changed_data) {
+        stat_flags |= ZIP_FL_UNCHANGED;
     }
 
-    if (zip_stat_index(srcza, srcidx, flags | ZIP_FL_UNCHANGED, &st) < 0) {
-        zip_error_set(error, ZIP_ER_INTERNAL, 0);
```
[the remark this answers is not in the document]
krnowak: Should it still be `ZIP_ER_INTERNAL` or should it rather copy the error stored in `srcza` into `error`, as `zip_stat_index` now could be calling third party code?

lib/zip_source_zip_new.c
```diff
@@ -126,6 +188,11 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
             return NULL;
         }
     }
+    else {
+        if ((src = _zip_source_window_new(srcza->entry[srcidx].source, 0, data_len, &st, &attributes, NULL, 0, error)) == NULL) {
```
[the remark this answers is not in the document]
krnowak: Another reason for wrapping it into window source is to make it possible to open `srcza->entry[srcidx].source` multiple times via `zip_file_t` and read each of the `zip_file_t` independently, because window source does a seek before each read.

man/zip_source_function.mdoc
```diff
@@ -109,6 +109,12 @@ Must additionally support
 .Dv ZIP_SOURCE_TELL_WRITE ,
 and
 .Dv ZIP_SOURCE_REMOVE .
+.Pp
+On top of the above, supporting a pseudo-command
+.Dv ZIP_SOURCE_SUPPORTS_REOPEN
+will allow calling
+.Fn zip_fopen
+without the ZIP_FL_UNCHANGED flag to read the modified data.
```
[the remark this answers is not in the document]
krnowak: Sorry, I don't understand this. You could read the data multiple times without this PR - it was possible because window source takes care of repositioning the underlying source by seeking it to the desired offset before doing a read. The catch was that it was only possible for unmodified data, and since unmodified data is backed by a source that implements seeking, reading data multiple times always worked.

lib/zip_source_zip_new.c
```diff
@@ -69,21 +87,37 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
...
-    if ((start > 0 || len > 0) && (start + len < start || start + len > st.size)) {
+    if ((start > 0 || len > 0) && ((start + len < start) || (have_size && (start + len > st.size)))) {
         zip_error_set(error, ZIP_ER_INVAL, 0);
         return NULL;
     }
 
     if (len == 0) {
-        len = st.size - start;
+        if (have_size) {
+            data_len = (zip_int64_t)(st.size - start);
```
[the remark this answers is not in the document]
krnowak: Done.

lib/zip_source_zip_new.c
```diff
@@ -100,24 +134,52 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
...
     }
 
     if (partial_data && !needs_decrypt && !needs_decompress) {
         struct zip_stat st2;
+        zip_t *source_archive;
+        zip_uint64_t source_index;
+
+        src = srcza->src;
+        source_archive = srcza;
+        source_index = srcidx;
```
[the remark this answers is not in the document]
krnowak: Done.

lib/zip_source_zip_new.c
```diff
@@ -100,24 +134,52 @@ zip_source_t *_zip_source_zip_new(zip_t *srcza, zip_uint64_t srcidx, zip_flags_t
     }
     _zip_file_attributes_from_dirent(&attributes, de);
 
-    if (st.comp_size == 0) {
+    have_comp_size = (st.valid & ZIP_STAT_COMP_SIZE) != 0;
+    if (compressed) {
+        if (have_comp_size && st.comp_size == 0) {
+            return zip_source_buffer_with_attributes_create(NULL, 0, 0, &attributes, error);
```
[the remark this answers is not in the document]
krnowak: Done.

lib/zip_source_zip_new.c
```diff
@@ -39,24 +39,42 @@
...
+    if ((flags & ZIP_FL_UNCHANGED) == 0) {
+        zip_entry_t *entry = srcza->entry + srcidx;
+        if (ZIP_ENTRY_DATA_CHANGED(entry)) {
+            if (!zip_source_supports_reopen(entry->source)) {
+                zip_error_set(error, ZIP_ER_CHANGED, 0);
+                return NULL;
+            }
+
+            changed_data = true;
+        } else if (entry->deleted) {
```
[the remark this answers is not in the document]
krnowak: Done.

regress/fopen_multiple.c
```diff
@@ -0,0 +1,197 @@
+/*
```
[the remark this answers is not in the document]
krnowak: Done, implemented fopen and fread commands in ziptool-regress and implemented the test in terms of those functions.

krnowak:
Updated. The only review issue left to address is the one about documentation, but I had some doubts about that as commented above. Other unresolved issues are solved, but had a comment or question about it anyway.

Implementing `cat` in terms of sources instead of files uncovered an issue with reading encrypted files (`decrypt-correct-password-pkware-2.test` was failing), hopefully the fix I have made is correct.

lib/zip_fclose.c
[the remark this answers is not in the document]
dillof: You're right, we need to keep track which references called open the source. We will split the source struct in two parts, the actual source and a reference. But we can do this after the merge.

lib/zip_source_zip.c
```diff
@@ -49,8 +51,14 @@ ZIP_EXTERN zip_source_t *zip_source_zip_create(zip_t *srcza, zip_uint64_t srcidx
...
-    
-    return _zip_source_zip_new(srcza, srcidx, flags, start, (zip_uint64_t)len, NULL, error);
+
+    password = srcza->default_password;
+
+    if (password != NULL && password[0] == '\0') {
+        password = NULL;
+    }
+
+    return _zip_source_zip_new(srcza, srcidx, flags, start, (zip_uint64_t)len, password, error);
```
dillof, line 61: `zip_set_default_password` already turns an empty password into `NULL`, so the check here is redundant. Just pass `srcza->default_password` to `_zip_source_zip_new`.

dillof (approved):

lib/zip_source_zip_new.c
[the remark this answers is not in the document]
dillof: We'll do this post merge.

dillof:
Sorry for letting this sit for so long. We finally merged it.

You did excellent work, thank you.

PR #305, opened 2022-06-21
Added python bindings

KOLANICH:
Currently they are completely untested. Under "untested" I mean they lack automatic tests. While I have used some subset of the bindings in some my software. I guess the easiest way to test it is to create a drop-in replacement for `zipfile` library of python standard library and just utilize their test suite, then cover the rest with specific tests.

Also they contain some bullshit within the code dealing with memory management of `zip_source`s: the docs prohibits calling `zip_source_free` at all, but since `source_keep`/`source_free` look like refcounting functions (IDK if I need them) I don't believe it should be implemented like that and probably `recordLeak` should be removed.

Also the wrappers for callbacks is not yet implemented.

### DIFF
--- a/python/.editorconfig
+++ b/python/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = tab
+indent_size = 4
+insert_final_newline = true
+end_of_line = lf
+
+[*.{yml,yaml}]
+indent_style = space
+indent_size = 2
--- a/python/.gitignore
+++ b/python/.gitignore
@@ -1,0 +1,11 @@
+__pycache__
+*.py[co]
+/*.egg-info
+*.srctrlbm
+*.srctrldb
+build
+dist
+.eggs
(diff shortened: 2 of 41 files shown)


dillof:
While this sounds interesting, it is best kept as a separate project, like all other language bindings for libzip. 

We will gladly add a link to it on the libzip web page.

dillof:
If you only use the official API, that will remain stable, so a close coupling is not necessary. That is precisely why maintaining backwards compatibility is such a high priority for us.

Also, it is quite assumptious to ask us to take over maintainership of your project. We don't have the time or motivation to do so.

KOLANICH:
The repo of the bindings is now situated here: https://github.com/KOLANICH-libs/libzip.py . @dillof, could you, please, review if I use the API right within them and within the tutorial? Also, if I had replaced a file within an archive, reading it would result in a crash. Do I understand right that I'm unallowed to read any deleted/replaced files within the archive?

PR #309, opened 2022-07-08
Address potential buffer overrun issues

samsappenfield:
Changes:

- Use more secure functions (e.g. *_s functions) for Microsoft CRT implementations that support secure libraries.
- Fixed possible buffer overrun issues by ensuring that allocation’s inputs don’t suffer from integer overflow or truncation (or just add a cast when the target variable cannot be truncated).

### DIFF
--- a/lib/zip_dirent.c
+++ b/lib/zip_dirent.c
@@ -1089,13 +1089,20 @@ _zip_get_dirent(zip_t *za, zip_uint64_t idx, zip_flags_t flags, zip_error_t *err
 
 void
 _zip_u2d_time(time_t intime, zip_uint16_t *dtime, zip_uint16_t *ddate) {
-    struct tm *tpm;
+    struct tm *tpm = NULL;
 
 #ifdef HAVE_LOCALTIME_R
     struct tm tm;
     tpm = localtime_r(&intime, &tm);
 #else
+#if __STDC_WANT_SECURE_LIB__
+    struct tm tm;
+    errno_t err = localtime_s(&tm, &intime);
+    if (err == 0)
+      tpm = &tm;
+#else
     tpm = localtime(&intime);
+#endif // __STDC_WANT_SECURE_LIB__
 #endif
     if (tpm == NULL) {
         /* if localtime() fails, return an arbitrary date (1980-01-01 00:00:00) */
--- a/lib/zip_source_buffer.c
+++ b/lib/zip_source_buffer.c
@@ -225,8 +225,10 @@ read_data(void *state, void *data, zip_uint64_t len, zip_source_cmd_t cmd) {
             zip_error_set(&ctx->error, ZIP_ER_INVAL, 0);
             return -1;
         }
-
-        (void)memcpy_s(data, len, &ctx->attributes, sizeof(ctx->attributes));
+        if (len != (size_t)len) {
+            return 0;
+        }
+        (void)memcpy_s(data, (size_t)len, &ctx->attributes, sizeof(ctx->attributes));
 
         return sizeof(ctx->attributes);
     }
@@ -538,7 +540,10 @@ buffer_read(buffer_t *buffer, zip_uint8_t *data, zip_uint64_t length) {
     while (n < length) {
         zip_uint64_t left = ZIP_MIN(length - n, buffer->fragments[i].length - fragment_offset);
 
-        (void)memcpy_s(data + n, length - n, buffer->fragments[i].data + fragment_offset, left);
+        if (length - n != (size_t)(length - n) || left != (size_t)left) {
+            return 0;
+        }
+        (void)memcpy_s(data + n, (size_t)(length - n), buffer->fragments[i].data + fragment_offset, (size_t)left);
 
         if (left == buffer->fragments[i].length - fragment_offset) {
             i++;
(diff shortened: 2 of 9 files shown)


samsappenfield:
Microsoft is now requiring '/sdl' flag for all internal products and the '/sdl' flag on the Microsoft toolchain requires the *_s functions. https://docs.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks?view=msvc-170 

The *_s functions are just versions of the original functions with security enhancements.
https://docs.microsoft.com/en-us/cpp/c-runtime-library/security-features-in-the-crt?view=msvc-170

dillof:
I see, so this is worth supporting.

I see three possibilities:

1. Keep using the legacy functions in libzip code and map them to their _s counterparts via defines. This kind of defeats the point, since they don't have the relevant information for buffer sizes.
2. Use the _s functions in libzip and add a compatibility layer for platforms without SDL.
    a. Map them via defines.
    b. Implement them, including their additional checks.
3. Create a more generic abstraction to use in libzip and implement that in terms of the _s or legacy functions, depending on availability. This should also handle similar enhancements on other platforms.

The main libzip source should use one set of functions constantly and remain readable, with a minimum of `#ifdef`s.
 
To decide which is best, I would need to know which functions have a _s counterpart, and what the additional arguments are. Do you have a list of all affected functions and their _s counterparts?

samsappenfield:
I do not have a complete list but here are the main ones that showed up for us as we looked through this repo with links to documentation for their _s counterparts: [memcpy](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/memcpy-s-wmemcpy-s?view=msvc-170), [strcpy](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strcpy-s-wcscpy-s-mbscpy-s?view=msvc-170), [sprintf](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/sprintf-s-sprintf-s-l-swprintf-s-swprintf-s-l?view=msvc-170), [strerror](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strerror-s-strerror-s-wcserror-s-wcserror-s?view=msvc-170), 
and [_snwprintf](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-s-snprintf-s-l-snwprintf-s-snwprintf-s-l?view=msvc-170)

lib/zip_dirent.c
```diff
@@ -219,11 +219,7 @@ _zip_dirent_clone(const zip_dirent_t *sde) {
         return NULL;
 
     if (sde)
-#if __STDC_WANT_SECURE_LIB__
-        memcpy_s(tde, sizeof(*sde), sde, sizeof(*sde));
-#else
-        memcpy(tde, sde, sizeof(*sde));
-#endif
+        _zip_memcpy(tde, sizeof(*sde), sde, sizeof(*sde));
```
0-wiz-0, line 222: Should that be `_zip_memcpy(tde, sizeof(*tde), sde, sizeof(*sde));`?

lib/zip_source_buffer.c
```diff
@@ -159,11 +160,7 @@ zip_source_buffer_fragment_with_attributes_create(const zip_buffer_fragment_t *f
     ctx->out = NULL;
     ctx->mtime = time(NULL);
     if (attributes) {
-#if __STDC_WANT_SECURE_LIB__
-        memcpy_s(&ctx->attributes, sizeof(ctx->attributes), attributes, sizeof(ctx->attributes));
-#else
-        memcpy(&ctx->attributes, attributes, sizeof(ctx->attributes));
-#endif
+        _zip_memcpy(&ctx->attributes, sizeof(ctx->attributes), attributes, sizeof(ctx->attributes));
```
0-wiz-0, line 163: Should that be `_zip_memcpy(&ctx->attributes, sizeof(ctx->attributes), attributes, sizeof(*attributes));`?

lib/zip_source_buffer.c
```diff
@@ -230,11 +227,7 @@ read_data(void *state, void *data, zip_uint64_t len, zip_source_cmd_t cmd) {
             return -1;
         }
 
-#if __STDC_WANT_SECURE_LIB__
-        memcpy_s(data, sizeof(ctx->attributes), &ctx->attributes, sizeof(ctx->attributes));
-#else
-        memcpy(data, &ctx->attributes, sizeof(ctx->attributes));
-#endif
+        _zip_memcpy(data, sizeof(ctx->attributes), &ctx->attributes, sizeof(ctx->attributes));
```
0-wiz-0, line 230: Should this be `_zip_memcpy(data, sizeof(*data), &ctx->attributes, sizeof(ctx->attributes));`?

0-wiz-0 (commented):
All but one calls to `_zip_memcpy` pass the _same_ length argument twice.
How does that improve security in any way?

dillof:
Since these secure functions are part of ISO C 11, we decided to use them and add compatibility defines for systems without them.

Could you please check if we still use functions that are forbidden?

samsappenfield:
We also need localtime_s and strcpy_s. We also do not allow fdopen and some of the values need to be cast as size_t. Thank you!

dillof:
We no longer use `localtime` or `strcpy`. You can disable the use of `fdopen` by running `cmake` with `-DENABLE_FDOPEN=OFF`.

Where do we need casts to `size_t`?

samsappenfield:
shelving for now

PR #319, opened 2022-08-06
Support mbedtls3 discovery

cryi:
MbedTLS moved version string to `build_info.h`. MbedTLS 3 API wise should maintain compatibility - at least I did not notice any issue.


0-wiz-0:
Thanks!

PR #324, opened 2022-08-22
Use win32 sources with mingw

cryi:
Right now in `MINGW` environment libzip uses linux sources and compilation fails.


cryi:
It is easier to achieve same when building as extarnal project within cmake and with flag 
`-DWIN32:BOOL=${MINGW}`. With above patch I had hard time to configure libzip as cmake subdirectory. Unix configuration wasn't detected properly. So closing for now.

PR #337, opened 2022-11-24
Implement zip_source_buffer_set_compression.

### LINKED ISSUE #292: Support pre compressed files
**Description**

In some cases it would be useful to support pre compressed files being added to zip without compressing again, but at the same time allow the decompression to still work on open. The reverse already is possible in libzip, if i want to extract the compressed data directly i can use something like `zip_fopen_index(zip_, index_, ZIP_FL_COMPRESSED)`

**Solution**
One option would be to leverage the compression flags of on zip_set_file_compression.  
`zip_set_file_compression(zip_, index,  ZIP_CM_STORE, ZIP_CM_DEFLATE);`
For example i have a pre deflated file that i want to just store, but i want to make sure the index reports that the content is deflated when opening. 

**Describe alternatives you've considered**
I looked through the code to see if there were any existing hooks and found nothing that stood out.

**Additional context**

arcanis:
This diff makes it possible to define a compression method for sources created by `zip_source_buffer_create`.

Closes #292


examples/raw-compression.c
```diff
@@ -0,0 +1,91 @@
...
+
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+
+#include <zip.h>
+
+char const comp_data[] = "\xcb\x48\xcd\xc9\xc9\x57\x28\xcf\x2f\xca\x49\x51\xb0\xd2\xe4\x02\x00";
```
arcanis, line 42: `"Hello world :)\n"`

dillof:
While the provided functionality is good, the approach is too specific and doesn’t fit cleanly in libzip’s architecture.

This functionality will be easy to implement cleanly once we make layered sources part of the public API, which we intend to do soon.

PR #342, opened 2022-12-18
Fix documentation for zip_source_callback

tsdgeos:
Add ZIP_SOURCE_FREE to the list of commands a read only source has to support


dillof:
It is not technically required, but we've mentioned it in the overview section.

PR #343, opened 2022-12-20
Drop unused member.

krnowak:
Looking into the source/reference split and noticed an unused member.


krnowak:
The appveyor failure seems to be unrelated, please see https://github.com/nih-at/libzip/pull/345.

PR #344, opened 2022-12-20
Factor our realloc handling to a separate function

krnowak:
While working on source/reference split, I added another growable array that held the references, which meant another piece of code doing overflow checks and invoking realloc, so I decided to factor that out to a single place. While at it, I changed one size member to use `zip_uint64_t`.

This also contains a small behavior change in one place, please see the comment in the code.

### DIFF
--- a/lib/zip_realloc.c
+++ b/lib/zip_realloc.c
@@ -1,6 +1,6 @@
 /*
-  zip_add_entry.c -- create and init struct zip_entry
-  Copyright (C) 1999-2021 Dieter Baron and Thomas Klausner
+  zip_realloc.c -- reallocate with additional elements
+  Copyright (C) 2009-2022 Dieter Baron and Thomas Klausner
 
   This file is part of libzip, a library to manipulate ZIP archives.
   The authors can be contacted at <info@libzip.org>
@@ -31,36 +31,45 @@
   IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
-
 #include <stdlib.h>
 
 #include "zipint.h"
 
+bool zip_realloc_implementation(void **inout_memory, zip_uint64_t *inout_alloced, zip_uint64_t element_size, zip_uint64_t additional_elements, zip_error_t *error) {
+    void *old_memory = *inout_memory;
+    zip_uint64_t old_alloced = *inout_alloced;
+    zip_uint64_t new_alloced;
+    size_t new_size;
+    void *new_memory;
 
-/* NOTE: Signed due to -1 on error.  See zip_add.c for more details. */
-
-zip_int64_t
-_zip_add_entry(zip_t *za) {
-    zip_uint64_t idx;
-
-    if (za->nentry + 1 >= za->nentry_alloc) {
-        zip_uint64_t additional_entries = 2 * za->nentry_alloc;
-
-        if (additional_entries < 16) {
-            additional_entries = 16;
-        }
-        else if (additional_entries > 1024) {
-            additional_entries = 1024;
-        }
-
-        if (!zip_realloc(&za->entry, &za->nentry_alloc, sizeof(struct zip_entry), additional_entries, &za->error)) {
-            return -1;
-        }
+    if (additional_elements == 0) {
+        return true;
     }
 
-    idx = za->nentry++;
+    if (old_alloced + additional_elements < old_alloced) {
+        zip_error_set(error, ZIP_ER_MEMORY, 0);
+        return false;
+    }
 
-    _zip_entry_init(za->entry + idx);
+    new_alloced = old_alloced + additional_elements;
 
-    return (zip_int64_t)idx;
+    if (new_alloced > SIZE_MAX / element_size) {
+        zip_error_set(error, ZIP_ER_MEMORY, 0);
+        return false;
+    }
+
+    /* Cast explicitly to size_t to quiet msvc complaints about
+       possible loss of data. At this point we know that the result of
+       "new_alloced * element_size" should fit into size_t. */
+    new_size = (size_t)(new_alloced * element_size);
+
+    if ((new_memory = realloc(old_memory, new_size)) == NULL) {
+        zip_error_set(error, ZIP_ER_MEMORY, 0);
+        return false;
+    }
+
+    *inout_memory = new_memory;
+    *inout_alloced = new_alloced;
+
+    return true;
 }
--- a/lib/zip_source_buffer.c
+++ b/lib/zip_source_buffer.c
@@ -427,32 +427,31 @@ buffer_free(buffer_t *buffer) {
 
 static bool
 buffer_grow_fragments(buffer_t *buffer, zip_uint64_t capacity, zip_error_t *error) {
-    zip_buffer_fragment_t *fragments;
-    zip_uint64_t *offsets;
+    zip_uint64_t tmp_fragments_capacity, tmp_offsets_capacity, additional_elements;
 
-    if (capacity < buffer->fragments_capacity) {
+    if (capacity <= buffer->fragments_capacity) {
         return true;
     }
 
-    zip_uint64_t fragments_size = sizeof(buffer->fragments[0]) * capacity;
-    zip_uint64_t offsets_size = sizeof(buffer->fragment_offsets[0]) * (capacity + 1);
-
-    if (capacity == ZIP_UINT64_MAX || fragments_size < capacity || fragments_size > SIZE_MAX|| offsets_size < capacity || offsets_size > SIZE_MAX) {
-        zip_error_set(error, ZIP_ER_MEMORY, 0);
+    tmp_fragments_capacity = buffer->fragments_capacity;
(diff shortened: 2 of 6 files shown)


lib/zip_source_buffer.c
```diff
@@ -429,30 +429,32 @@ static bool
 buffer_grow_fragments(buffer_t *buffer, zip_uint64_t capacity, zip_error_t *error) {
     zip_buffer_fragment_t *fragments;
     zip_uint64_t *offsets;
+    int result;
+    zip_uint64_t tmp_fragments_capacity, tmp_offsets_capacity, additional_elements;
 
-    if (capacity < buffer->fragments_capacity) {
+    if (capacity <= buffer->fragments_capacity) {
```
krnowak, line 435: This should be fine, IMO, requested capacity is still there if it's equal to the already existing one.

dillof: Yes, that was a bug, your version is correct.

lib/zipint.h
```diff
@@ -487,9 +487,12 @@ typedef struct _zip_pkware_keys zip_pkware_keys_t;
 #endif
 #endif
 
+#define _zip_realloc(inout_memory, inout_alloced, element_size, additional_elements) (_zip_realloc)((void **)inout_memory, inout_alloced, element_size, additional_elements)
+int (_zip_realloc)(void **inout_memory, zip_uint64_t *inout_alloced, zip_uint64_t element_size, zip_uint64_t additional_elements);
```
dillof, line 491: Please use a different name for macro and function (like `zip_realloc_implementation` for the function).

Also, since some C compilers warn about symbols beginning with `_` as being reserved, please don't use that prefix for new internal symbols.

lib/zip_source_buffer.c
```diff
@@ -429,30 +429,32 @@ static bool
...
-
-    if (capacity == ZIP_UINT64_MAX || fragments_size < capacity || fragments_size > SIZE_MAX|| offsets_size < capacity || offsets_size > SIZE_MAX) {
-        zip_error_set(error, ZIP_ER_MEMORY, 0);
+    tmp_fragments_capacity = buffer->fragments_capacity;
+    additional_elements = capacity - buffer->fragments_capacity;
+    result = _zip_realloc(&buffer->fragments, &tmp_fragments_capacity, sizeof(buffer->fragments[0]), additional_elements);
+    if (result != ZIP_ER_OK) {
+        zip_error_set(error, result, 0);
         return false;
     }
```
dillof, line 446: Why do you need the temporary variables? Why not passing in pointers to the members of `buffer` directly to `zip_realloc`.

I know that two arrays need to be resized, but each has its own capacity member, and if `zip_realloc` succeed for the first one, that is already resized and doesn't need to be resized again. So it would probably be best to decouple these two sections completely: If `capacity` > `buffer->fragments_capacity`, resize `buffer->fragments`. If `capacity + 1 > buffer->offsets_capacity`, resize `buffer->fragment_offsets`.

Also, `capacity + 1` needs an overflow check.

dillof (changes requested):
Thanks, that's an excellent idea, which should make code simpler and easier to understand.

I have some suggestions to make it fit better with the conventions used by libzip, see below.

(I guess we should start a style guide.)

lib/zip_realloc.c
```diff
@@ -0,0 +1,66 @@
...
+int
+(_zip_realloc)(void **inout_memory, zip_uint64_t *inout_alloced, zip_uint64_t element_size, zip_uint64_t additional_elements) {
+    void *old_memory = *inout_memory;
+    zip_uint64_t old_alloced = *inout_alloced;
+    zip_uint64_t new_alloced;
+    size_t new_size;
+    void *new_memory;
+
+    if (additional_elements == 0)
+        additional_elements = 1;
```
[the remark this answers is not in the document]
krnowak: Honestly? I don't remember. :) I wrote this function back in July or August, when working on that streams branch. Just borrowed that code from there. I probably should have documented this case. Anyway, will add the early return here instead.

lib/zip_source_buffer.c
krnowak: > Why do you need the temporary variables? Why not passing in pointers to the members of `buffer` directly to `zip_realloc`.
> 
> I know that two arrays need to be resized, but each has its own capacity member, and if `zip_realloc` succeed for the first one, that is already resized and doesn't need to be resized again. So it would probably be best to decouple these two sections completely: If `capacity` > `buffer->fragments_capacity`, resize `buffer->fragments`. If `capacity + 1 > buffer->offsets_capacity`, resize `buffer->fragment_offsets`.

The reason here is that those two arrays have actually a single capacity member, `fragments_capacity`. There's no `offsets_capacity`. So I'd like to update `fragments_capacity` after both reallocs succeeded.

> 
> Also, `capacity + 1` needs an overflow check.

Will look into it.

dillof: I see, I've misread the code, sorry. Your solution is correct.

lib/zip_realloc.c
```diff
@@ -0,0 +1,66 @@
...
+  IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+
+#include <stdlib.h>
+
+#include "zipint.h"
+
+int
+(_zip_realloc)(void **inout_memory, zip_uint64_t *inout_alloced, zip_uint64_t element_size, zip_uint64_t additional_elements) {
```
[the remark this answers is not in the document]
krnowak: > Error propagation in libzip is done via a `zip_error_t *error` argument, usually the last argument of the function. The function fills in `error` via `zip_error_set()` and signals that an error occurred via a special return value (often `-1` or `false`).

This made the calling code even simpler, thanks.

> Also, please put the return type and function prototype on the same line. (This is our new style; we haven't converted all of libzip to this new style yet.)

Alright, done.

lib/zip_realloc.c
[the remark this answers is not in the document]
krnowak: Done.

lib/zip_realloc.c
```diff
@@ -0,0 +1,66 @@
...
+    void *old_memory = *inout_memory;
+    zip_uint64_t old_alloced = *inout_alloced;
+    zip_uint64_t new_alloced;
+    size_t new_size;
+    void *new_memory;
+
+    if (additional_elements == 0)
+        additional_elements = 1;
+
+    if (old_alloced > ZIP_UINT64_MAX - additional_elements)
```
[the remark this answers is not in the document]
krnowak: Done. I kept the multiplication overflow check as is, though.

lib/zipint.h
krnowak: Done.

lib/zip_source_buffer.c
krnowak: > Also, `capacity + 1` needs an overflow check.

Added a comment about why this overflow is not possible.

lib/zipint.h
```diff
@@ -486,6 +486,8 @@ typedef struct _zip_pkware_keys zip_pkware_keys_t;
 #endif
 #endif
 
+#define zip_realloc(inout_memory, inout_alloced, element_size, additional_elements, error) zip_realloc_implementation((void **)inout_memory, inout_alloced, element_size, additional_elements, error)
```
krnowak, line 489: I was thinking about dropping the `element_size` parameter from the macro and make the macro to pass `(sizeof(**(inout_memory)))` instead, but this is probably too much magic.

lib/zip_realloc.c
```diff
@@ -0,0 +1,85 @@
...
+       of SIZE_MAX being greater than ZIP_UINT64_MAX. */
+#if ZIP_UINT64_MAX > SIZE_MAX
+    max = SIZE_MAX;
+#else
+    max = ZIP_UINT64_MAX;
+#endif
+    if (new_alloced > max / element_size) {
+        zip_error_set(error, ZIP_ER_MEMORY, 0);
+        return false;
+    }
```
krnowak, line 69: Not sure if this fiddling with `max` is necessary, please let me know what you think. In general switching between `size_t` and `uint64` seems to be icky, especially on Windows platforms.

dillof: You can always use `SIZE_MAX` here: `new_allocated * element_size` must fit in a `size_t`.

lib/zip_source_buffer.c
dillof: Ah, I see, I just didn't fully understand the comment. This seems clearer to me, feel free to reword it if you disagree:

This is technically not correct if buffer->fragment_capacity is 0, but results in the correct required new capacity for buffer->fragment_offsets in all cases (one element more than for buffer->fragments). Also, since we added additional_elements (which is 1 or more) to buffer->fragments_capacity in the previous realloc, we know it can't overflow.

lib/zip_realloc.c
krnowak: Done.

lib/zip_source_buffer.c
krnowak: Used your wording, thanks.

PR #349, opened 2023-01-19
Resolve optional transitive dependencies in libzip-config.cmake.

mmassing:
This is required when building libzip as a static library.


0-wiz-0:
Thank you, merged!

PR #370, opened 2023-03-29
CMake: Use upstream names for zstd

FtZPetruska:
The upstream zstd package uses different naming and targets than what the current Find module uses. This leads to limitations such as:
- upstream config being ignored when setting `CMAKE_FIND_PACKAGE_PREFER_CONFIG` to `ON`
- passing the upstream config through `Zstd_DIR=/path/to/lib/cmake/zstd/zstdConfig.cmake` gives incompatible targets.

Upstream the package uses the following names:
- `zstd`, all lowercase, for the package name
- `zstd::libzstd_shared` for the shared library
- `zstd::libzstd_static` for the static library

As there is no "generic" target (a target that would be dynamically set to the `_shared` or `_static` variant depending on which is found), both need to be searched in the Find module.

Additional steps must be taken to handle some Windows specificities:
- For MSVC, the file names are `zstd.dll`, the import library `zstd.lib`, and the static library `zstd_static.lib`.
- When using vcpkg, since the builds are static-only or shared-only, they remove the `_static` suffix on the static library.
- For MSYS2 (mingw), the file names are `libzstd.dll`, the import library `libzstd.dll.a`, and the static library `libzstd.a`.

To account for the vcpkg-specific behaviour, I used the following logic (for MSVC):
- If a DLL is found, the import lib should be `zstd.lib` and there may be a `zstd_static.lib` static lib.
- If no DLL is found, there is no import lib and both `zstd.lib` and `zstd_static.lib` may be the static lib.
- Then for the calls to `find_library`, I overwrite `CMAKE_FIND_LIBRARY_SUFFIXES` to exclude `.lib` and instead search the MSVC libraries by full names.

---

Here are the following configurations I tested, All the tests were conducted with and without the `CMAKE_FIND_PACKAGE_PREFER_CONFIG` option:

### Windows

- MSVC (no pkg-config)
  - Using a custom static-only build of zstd
  - Using a custom static+shared build of zstd
  - Using vcpkg with the triplet x64-windows (shared-only)
  - Using vcpkg with the triplet x64-windows-static (static-only)
- MSYS2, using the package manager version (static+shared, has pkg-config)

### macOS

- Using Homebrew-provided package (static+shared)
- Using vcpkg with the triplet arm64-osx (static-only)


FtZPetruska:
> I'm a bit confused by what you mean with 'The upstream zstd package' name because the versions I know don't install any cmake files.

Seems like zstd has a variety of ways to be built, with Visual Studio solutions, CMake, meson, Makefile, or simple shell scripts. The CMake config is only installed when zstd is built with CMake, which is the case when using Homebrew or vcpkg. For Linux distributions such as Debian, they tend to prefer using Makefiles which mean they do not ship the CMake config.

Hopefully this clears it up!

eli-schwartz:
> For Linux distributions such as Debian, they tend to prefer using Makefiles which mean they do not ship the CMake config.

Canonically speaking, the zstd project considers the Makefiles to be the only "official" solution. All other methods are unofficial build systems and their results are not canonical.

The pkg-config files are installed by all build systems (if you don't count VS as a real build system) which means that:
- it should probably be preferred when checking for available dependencies
- you don't actually need to mangle flags when producing your own pkg-config files, simply use the "Requires" feature of pkg-config

PR #374, opened 2023-04-13
Regenerate zip_err_str.c only when needed

chaoticgd:
I'm currently using libzip as a static library in one of my projects by including it as a git submodule and [using the add_subdirectory command in cmake](https://github.com/chaoticgd/wrench/blob/0f0d4e77a6070e8401ba99065ef948260e1eec2d/thirdparty/CMakeLists.txt#L95). Every time I configure the parent project zip_err_str.c is regenerated, meaning libzip and all targets that use libzip and so on have to be relinked. None of the other libraries I'm using seem to do this, which made me suspect this was a problem with libzip's build scripts.

This pull requests modifies the lib/CMakeLists.txt file so that zip_err_str.c is generated at build time instead of configure time and only if necessary. The logic that generates the file has been moved to its own cmake script which is invoked via add_custom_command. The dependency between libzip and this script [is automatically added by cmake](https://cmake.org/cmake/help/latest/prop_sf/GENERATED.html).

This seems a little complicated, so I considered just including the generated code in the repository instead, but I wanted to keep everything as close as possible to how it was working before. Another approach would've been to use preprocessor hacks, but I don't think that allows for generating more macros, so I ruled that out.

### DIFF
--- a/cmake/GenerateZipErrorStrings.cmake
+++ b/cmake/GenerateZipErrorStrings.cmake
@@ -1,0 +1,47 @@
+# create zip_err_str.c from zip.h and zipint.h
+file(READ ${PROJECT_SOURCE_DIR}/lib/zip.h zip_h)
+string(REGEX MATCHALL "#define ZIP_ER_([A-Z0-9_]+) ([0-9]+)[ \t]+/([-*0-9a-zA-Z, ']*)/" zip_h_err ${zip_h})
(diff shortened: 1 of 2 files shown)


chaoticgd:
Also thanks for this library! It's the sanest way I've found to read/write zip files from C/C++.

PR #392, opened 2023-07-12
cmake: bump minimum required to version 3.5.0

scribam:
This fixes the following warning when configuring with CMake 3.27+:
```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```


0-wiz-0:
3.4 is from late 2015, 3.5 is from early 2016 - sure, let's bump it.

PR #395, opened 2023-07-23
Fallback to srand/rand if srandom/random are missing

scribam:
From https://github.com/nih-at/libzip/pull/156
> 1. `srandom` and `random` are POSIX extensions not present on non-POSIX
>        platforms, replace them with `srand` and `rand` respectively. They
>        _might_ produce different results, but that shouldn't matter in the
>        context of the caller, which is the non-secure version

uwp and win32 already use these functions, we can unify all platforms with these changes

vitasdk (a sdk for playstation vita) has the same patch (https://github.com/vitasdk/packages/blob/master/libzip/random.patch) in order to build libzip and I would like to remove it :)


dillof:
IIRC, some systems have deprecation warning for `rand()`. I'd therefore prefer to have cmake check for the existence of `random()` and only fall back to `rand()` if it's missing.

Also, while reading the diff, I just noticed that we never set `seeded` to `true`. Could you please add this, so we avoid conflicts?

PR #401, opened 2023-08-23
Define WIN32_LEAN_AND_MEAN to build the zip library on Windows

scribam:
There are other files where windows.h is included but WIN32_LEAN_AND_MEAN is not defined. With this PR, WIN32_LEAN_AND_MEAN is defined consistently across the project

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -117,6 +117,7 @@ add_library(zip
 add_library(libzip::zip ALIAS zip)
 
 if(WIN32)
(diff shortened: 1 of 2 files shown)


dillof:
Why did you remove NOCRYPT? Are you sure that does not influence `<windows.h>` or `<bcrypt.h>`?

dillof:
Thanks for the explanation, makes sense.

0-wiz-0:
Thanks, merged.

PR #406, opened 2023-09-01
Add a toolchain for publishing to NuGet

Silver-Fang:
Not sure if maintainers are interested in this?


dengchj:
Maybe English is more common.

Silver-Fang:
> Maybe English is more common.

I'll deal with that. But what about this appveyor failure?

dillof:
> > Maybe English is more common.
> 
> I'll deal with that. But what about this appveyor failure?

Don't worry about that, one of the dependencies fail to install and we haven't had the time to sort that out yet.

0-wiz-0:
I see there is "libzip-c" in NuGet already, which seems quite up-to-date.
We do not add packaging information for packaging systems into our repository.
Thank you for packaging libzip!

Issue #377, opened 2023-05-02
Reference actions by commit SHA


gabibguti:
**Description**
Referencing actions by commit SHA in workflows to keep your repository safe against supply-chain attacks.
As for now, I didn't find this feature mentioned in TODO.md.

**Solution**
I'd like us to reference actions by commit SHA in workflows to ensure we are using an immutable version. Actions referenced by tags and branches are vulnerable to attacks, such as moving a tag to a malicious commit, pushing a malicious commit to the branch, or typosquatting.

While each reference has its advantages and disadvantages, GitHub acknowledges that [using commit SHAs is more reliable](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-shas), as does [Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies) security tool.

Currently, this repository uses actions referenced by tags and branches, such as `actions/upload-artifact@v1` and `google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master`. If you agree, I can open a PR to change these references to commit SHAs.

**Describe alternatives you've considered**
None.

**Additional context**
Hi again! I'm Gabriela, I recently suggested https://github.com/nih-at/libzip/issues/364, and I work on behalf of Google and the OpenSSF suggesting supply-chain security changes :)

0-wiz-0:
Hi again!
We've actually recently be talking about this internally.
Right now, we have not even tagged all workflows we use, but use master for some of them.
If we used the git hashes and/or tags everywhere, we'd have to monitor that new versions of the workflows are available.
How do other projects do that?

On the other hand, we don't see an effective attack potential here - we are using workflows for fuzzing and CI, not for creating releases. So we might just not get correct information about the build or fuzz status. What other possible problems do you see here?

gabibguti:
> If we used the git hashes and/or tags everywhere, we'd have to monitor that new versions of the workflows are available.
How do other projects do that?

You can use dependabot or renovatebot. These bots can be configured to update only workflows, and they can identify commit SHA + comments updates.

```
e.g.
uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # v3.1.1
updates to
uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
```

gabibguti:
[...]

PR #408, opened 2023-09-15
Reference actions by commit SHA

gabibguti:
Resolves https://github.com/nih-at/libzip/issues/377

This PR changes GitHub Actions being referenced by tags like `@v1` to `@{commit-SHA}`.
We're not updating `oss-fuzz` since the project architecture does not support referencing by commit SHA.

Also, I noticed the action `lukka/run-vcpkg` is being used in major 7, but already has a major 11, would you like me to try bumping that action?

### DIFF
--- a/.github/workflows/CIFuzz.yml
+++ b/.github/workflows/CIFuzz.yml
@@ -18,7 +18,7 @@ jobs:
         fuzz-seconds: 600
         dry-run: false
     - name: Upload Crash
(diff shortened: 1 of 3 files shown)


gabibguti:
To install renovatebot, configure it in this link: https://github.com/apps/renovate
You can follow these instructions to do the configuration: https://github.com/renovatebot/tutorial#part-1---installation

PR #409, opened 2023-09-22
I/O problems in fuzzers

nmlsg:
Hello,

I've tested your repository and noticed that it generates numerous .zip archives (specifically in the zip_read_file_fuzzer function) due to the absence of file deletion in certain if statements used for logging purposes.

For instance, in the code snippet:
```
if (source == NULL) {
    printf("failed to create source buffer. %s\n", zip_strerror(archive));
    return -1;
}
```
I suggest modifying it to:

```
if (source == NULL) {
    std::remove(path.c_str());
    printf("failed to create source buffer. %s\n", zip_strerror(archive));
    return -1;
}
```

This adjustment will ensure that any failed file creations are followed by the deletion of the associated files. Without this modification, multiple .zip archives may accumulate without being deleted.

Thank you for your attention to this matter.


0-wiz-0:
Thank you, merged.

PR #413, opened 2023-09-25
Add MACHO_*_VERSION when building the dylib on macOS

### LINKED ISSUE #405: change from autotools to cmake decreased library compatibility_version (macOS)
**Describe the Bug**
v1.3.2 was the last version of libzip that used autotools. On macOS, it installed libzip.5.dylib with a compatibility_version (c_v hereafter) of 6.0.0. Building v1.10.1 with cmake, it also installs the library libzip.5.dylib, but the c_v has been downgraded to 5.0.0. This means that anything built linking against v1.3.2 will fail to run if libzip is updated. The system will see that c_v has been decreased, throw an error, and exit.

**Expected Behavior**
c_v should be 6.0.0 or greater.

**To Reproduce**
This shows the output from `otool -L` for v1.3.2 and v1.10.1
```
$  otool -L /sw/lib/libzip.5.dylib ~/src/libzip-1.10.1/finkbuild/lib/libzip.5.dylib
/sw/lib/libzip.5.dylib:
	/sw/lib/libzip.5.dylib (compatibility version 6.0.0, current version 6.0.0)
	...
/Users/nieder/src/libzip-1.10.1/finkbuild/lib/libzip.5.dylib:
	@rpath/libzip.5.dylib (compatibility version 5.0.0, current version 5.5.0)
	...
```

**libzip Version**
v1.3.2 vs v1.10.1

**Operating System**
macOS 10.14.6, using Apple-clang, but this will affect any macOS system.

**Additional context**
The problem here arose because autotools/libtool defaulted to c_v always being SOVERSION+1. But cmake defaults to c_v=SOVERSION.

cmake provides the properties `MACHO_COMPATIBILITY_VERSION` and `MACHO_CURRENT_VERSION` to deal with updates to the API (such as addition of new symbols).
https://cmake.org/cmake/help/latest/prop_tgt/MACHO_COMPATIBILITY_VERSION.html

Also, this will only be an issue while SOVERSION=5. Once SOVERSION gets increased and thus the library gets a new install_name, then c_v can be anything new (for example 1 for being the first release of the new SOVERSION, or 6 by tracking SOVERSION).

The code in question:
https://github.com/nih-at/libzip/blob/c719428916b4d19e838f873b1a177b126a080d61/lib/CMakeLists.txt#L174-L176
Adding `MACHO_COMPATIBILITY_VERSION 6 MACHO_CURRENT_VERSION 6.5` yields this:
```
$ otool -L lib/libzip.5.dylib 
lib/libzip.5.dylib:
	@rpath/libzip.5.dylib (compatibility version 6.0.0, current version 6.5.0)
```
which means the new v1.10.1 will work with stuff that linked against the older library.

nieder:
Fixes #405

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -172,7 +172,8 @@ if(HAVE_CRYPTO)
 endif()
 
 if(SHARED_LIB_VERSIONNING)
-  set_target_properties(zip PROPERTIES VERSION 5.5 SOVERSION 5)
+  # MACHO_*_VERSION can be removed when SOVERSION gets increased. Cf #405
+  set_target_properties(zip PROPERTIES VERSION 5.5 SOVERSION 5 MACHO_CURRENT_VERSION 6.5 MACHO_COMPATIBILITY_VERSION 6)
 endif()
 
 target_link_libraries(zip PRIVATE ZLIB::ZLIB)


0-wiz-0:
Thank you!

PR #434, opened 2024-05-19
pkg-config: Add Cflags.private field for static linking

Biswa96:
This allows for the usage of $(pkgconf --static --cflags libzip) to produce
    the proper CFLAGS for static linking. Relies on pkgconf rather than pkg-config.

### DIFF
--- a/libzip.pc.in
+++ b/libzip.pc.in
@@ -12,3 +12,4 @@ Version: @PROJECT_VERSION@
 Libs: @PKG_CONFIG_RPATH@ -L${libdir} -lzip
 Libs.private: @LIBS@
 Cflags: -I${includedir}
+Cflags.private: -DZIP_STATIC


0-wiz-0:
`ZIP_STATIC` is already set in `zipconf.h`, which is installed by default and included from `zip.h`.
(Also, this patch unconditionally adds `-DZIP_STATIC` to the pkg-config file - this is wrong for non-static builds.)

Biswa96:
`-DZIP_STATIC` is added to CFLAGS if pkgconfig is called with `-static` option. So, non-static build will not be affected with this change.

Biswa96:
BTW, gcc 14 shows compiler error with function pointers in Windows. The full build log can be found https://github.com/msys2/MINGW-packages/pull/20943/checks#step:11:566 where the errors were downgraded to warnings.

0-wiz-0:
The GCC 14 warnings/errors can be fixed with casts. You could open a separate pull request for those. (We do not have Windows expertise and depend on contributions by knowledgeable Windows developers there.)

PR #436, opened 2024-05-23
Make the library compile when targeting GDK

sykhro:
GDK is the Microsoft framework for develping games for the Store and Xbox. It defines `WINAPI_PARTITION_GAMES` (like UWP does) and a build target, `Gaming.Desktop.x64`. It's like compiling for Windows but without a chunk of Win32.
If you are a licensed developer, you also get the build targets for Xbox. 
To get this platform working, you can have a look at the [GDK CMake examples](https://github.com/microsoft/Xbox-GDK-Samples/blob/main/Samples/Tools/CMakeGDKExample/grdk_toolchain.cmake).

This PR:
- Adds a CMake check to see if the compiler is targeting WINAPI_FAMILY_GAMES
  - This is done to ensure backward compatibility with older MSVC (which chokes on that macro) and WinSDK headers (WINAPI_FAMILY_GAMES exists as of Windows 10)
- Avoids calling `GetSecurityInfo()` as it's not defined in the API partition mentioned above

making libzip compile when targeting `Gaming.*` platforms.

### DIFF
--- a/lib/zip_source_file_win32_named.c
+++ b/lib/zip_source_file_win32_named.c
@@ -33,6 +33,13 @@
 
 #include "zip_source_file_win32.h"
 
+/* ACL is not available when targeting the games API partition */
+#if defined(WINAPI_FAMILY_PARTITION) && defined(WINAPI_PARTITION_GAMES)
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_GAMES)
+#define ACL_UNSUPPORTED
+#endif
+#endif
+
 static zip_int64_t _zip_win32_named_op_commit_write(zip_source_file_context_t *ctx);
 static zip_int64_t _zip_win32_named_op_create_temp_output(zip_source_file_context_t *ctx);
 static bool _zip_win32_named_op_open(zip_source_file_context_t *ctx);
@@ -110,7 +117,11 @@ _zip_win32_named_op_create_temp_output(zip_source_file_context_t *ctx) {
 
     if ((HANDLE)ctx->f != INVALID_HANDLE_VALUE && GetFileType((HANDLE)ctx->f) == FILE_TYPE_DISK) {
         si = DACL_SECURITY_INFORMATION | UNPROTECTED_DACL_SECURITY_INFORMATION;
+    #ifdef ACL_UNSUPPORTED
+        success = ERROR_NOT_SUPPORTED;
+    #else
         success = GetSecurityInfo((HANDLE)ctx->f, SE_FILE_OBJECT, si, NULL, NULL, &dacl, NULL, &psd);
+    #endif
         if (success == ERROR_SUCCESS) {
             sa.nLength = sizeof(SECURITY_ATTRIBUTES);
             sa.bInheritHandle = FALSE;


CMakeLists.txt
```diff
@@ -282,7 +282,24 @@ endif(MSVC)
...
   if(CMAKE_SYSTEM_NAME MATCHES WindowsPhone OR CMAKE_SYSTEM_NAME MATCHES WindowsStore)
     add_compile_definitions(MS_UWP)
-  endif(CMAKE_SYSTEM_NAME MATCHES WindowsPhone OR CMAKE_SYSTEM_NAME MATCHES WindowsStore)
+  else()
+    check_c_source_compiles(
+      "
+      #include <winapifamily.h>
+      
+      int main() {
+        #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_GAMES)
```
0-wiz-0, line 291: Could we check, in the one place where it's needed, instead:
```
#if defined(WINAPI_FAMILY_PARTITION) && defined(WINAPI_PARTITION_GAMES) && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_GAMES)
```
(avoiding the `MS_GAMES` symbol completely)?

0-wiz-0 (commented):
Thanks for the patch! Can you please comment on my suggestions and adapt the patches?

lib/zip_source_file_win32_named.c
```diff
@@ -108,6 +108,13 @@ _zip_win32_named_op_create_temp_output(zip_source_file_context_t *ctx) {
     char *tempname = NULL;
     size_t tempname_size = 0;
 
+    /* ACL is not available when targeting the games api partition */
+    #ifdef MS_GAMES
+    if ((HANDLE)ctx->f != INVALID_HANDLE_VALUE && GetFileType((HANDLE)ctx->f) == FILE_TYPE_DISK) {
+        si = DACL_SECURITY_INFORMATION | UNPROTECTED_DACL_SECURITY_INFORMATION;
```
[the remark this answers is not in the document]
sykhro: I wasn't sure if any code later on checked for it, I'll remove it

CMakeLists.txt
sykhro: Sure, I did it this way to mimic what was done for UWP and to avoid potential compat issues with older compilers. Happy to remove it

sykhro: (the very old MSVC preprocessor apparently doesn't like `WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_GAMES)`, do you support that config in this project?)

0-wiz-0: Even if you separately check for the two symbol parts first?
How old is "old MSVC preprocessor"?

sykhro: 2015? possibly earlier

0-wiz-0: And https://github.com/nih-at/libzip/pull/436#discussion_r1611471748 is not safe enough for that?

sykhro: Apparently yes. I was proposing the same approach for a patch on freetype and the reply I got was:
```
We cannot use  WINAPI_FAMILY_PARTITION(), because some other compilers might choke on it. Even old MSVC might not be happy. So we have to keep it simple for other compilers to understand.
```
and also
```
Apparently, WINAPI_FAMILY_GAMES is too new to use [...]
```
what are the supported configs for this project? The check you proposed should be okay all the way down to win8

0-wiz-0: I guess we need to split it in two lines:
```
#if defined(WINAPI_FAMILY_PARTITION) && defined(WINAPI_PARTITION_GAMES)
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_GAMES)
```

dillof: Maybe the problem isn't that it's not defined, but that it *is* defined and the older compilers can't handle it? In that case we would also need to check compiler versions.

0-wiz-0: I found [this](https://github.com/google/benchmark/issues/1345#issuecomment-1039657727) and thought it was a reasonable explanation.

dillof: Good, then splitting it into two lines as you suggested should work.

lib/zip_source_file_win32_named.c
```diff
@@ -110,7 +110,14 @@ _zip_win32_named_op_create_temp_output(zip_source_file_context_t *ctx) {
...
     if ((HANDLE)ctx->f != INVALID_HANDLE_VALUE && GetFileType((HANDLE)ctx->f) == FILE_TYPE_DISK) {
         si = DACL_SECURITY_INFORMATION | UNPROTECTED_DACL_SECURITY_INFORMATION;
+    #if defined(WINAPI_FAMILY_PARTITION) && defined(WINAPI_PARTITION_GAMES)
+    #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_GAMES)
+        /* ACL is not available when targeting the games API partition */
+        success = ERROR_NOT_SUPPORTED;
+    #else
         success = GetSecurityInfo((HANDLE)ctx->f, SE_FILE_OBJECT, si, NULL, NULL, &dacl, NULL, &psd);
+    #endif
+    #endif
```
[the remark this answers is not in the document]
sykhro: right, I have duplicated the line below for lack of better ideas

sykhro:
Done. It all seems like it works, will send more patches for runtime behaviour if I run into anything else

0-wiz-0:
Thank you, merged!

PR #446, opened 2024-07-24
Fix multiple initializations from incompatible pointer types

manisandro:
When building with GCC14 to MinGW, the build currently fails due to multiple initializations of incompatible pointer types such as

```
/builddir/build/BUILD/mingw-libzip-1.10.1-build/libzip-1.10.1/lib/zip_source_file_win32_utf16.c:46:5: error: initialization of 'void * (__attribute__((stdcall)) *)(const void *, DWORD,  DWORD,  struct _SECURITY_ATTRIBUTES *, DWORD,  DWORD,  void *)' {aka 'void * (__attribute__((stdcall)) *)(const void *, long unsigned int,  long unsigned int,  struct _SECURITY_ATTRIBUTES *, long unsigned int,  long unsigned int,  void *)'} from incompatible pointer type 'void * (__attribute__((stdcall)) *)(const char *, DWORD,  DWORD,  struct _SECURITY_ATTRIBUTES *, DWORD,  DWORD,  void *)' {aka 'void * (__attribute__((stdcall)) *)(const char *, long unsigned int,  long unsigned int,  struct _SECURITY_ATTRIBUTES *, long unsigned int,  long unsigned int,  void *)'} [-Wincompatible-pointer-types]
   46 |     utf16_create_file,
      |     ^~~~~~~~~~~~~~~~~
```

See also https://kojipkgs.fedoraproject.org//work/tasks/1551/120721551/build.log

### DIFF
--- a/lib/zip_source_file_win32_ansi.c
+++ b/lib/zip_source_file_win32_ansi.c
@@ -34,20 +34,26 @@
 #include "zip_source_file_win32.h"
 
 static char *ansi_allocate_tempname(const char *name, size_t extra_chars, size_t *lengthp);
+static HANDLE __stdcall ansi_create_file(const void *name, DWORD access, DWORD share_mode, PSECURITY_ATTRIBUTES security_attributes, DWORD creation_disposition, DWORD file_attributes, HANDLE template_file);
+static BOOL __stdcall ansi_delete_file(const void *name);
+static DWORD __stdcall ansi_get_file_attributes(const void *name);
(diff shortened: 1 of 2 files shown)


0-wiz-0:
Thank you, merged!
Can we now remove the block starting from
After https://github.com/nih-at/libzip/pull/446 the whole block starting from https://github.com/nih-at/libzip/blob/105e7a8380e45dc593173c49da620fe06d965cbb/lib/zip_source_file_win32.h#L76 should not be necessary any longer - can you please confirm that it can now be removed?
(and the usages of `DONT_WARN_INCOMPATIBLE_FN_PTR_BEGIN/END`)

PR #450, opened 2024-08-14
Fix zip_open failure with data-deduplicated file (Windows)

### LINKED ISSUE #449: zip_open fails with a file on the filesystem with data deduplication turned on (Windows)
**Describe the Bug**
[zip_open](https://libzip.org/documentation/zip_open.html) fails with a file on the Windows Server which has [data deduplication](https://learn.microsoft.com/en-us/windows-server/storage/data-deduplication/understand) turned on.  

**Expected Behavior**
The data deduplication should be transparent to the client app according to Microsoft.

"Optimization should not change access semantics Users and applications that access data on an optimized volume are completely unaware that the files they are accessing have been deduplicated."

Therefore, `zip_open` is expected to succeed.

**Observed Behavior**
`zip_open` fails with error code 28 (ZIP_ER_OPNOTSUPP)

**To Reproduce**
Add a zip file to the file system with data deduplication turned on, and try opening the file with `zip_open`.

**libzip Version**
v1.10.1 as well as the current tip of `main`

**Operating System**
Windows

**Test Files**
Any zip file

**Additional context**
This is a similar issue to #382 since in both cases, the file's attribute contains `FILE_ATTRIBUTE_REPARSE_POINT`.
I'll follow up with a PR to address this issue.

Thank you!!

ktakahashimtb:
Data-deduplicated files should be treated as regular files since this feature is designed to be transparent to the filesystem client.  It requires an additional check on the reparse point tag after file attributes reveals it uses a reparse point. 
Fixes #449

### DIFF
--- a/lib/zip_source_file_win32.h
+++ b/lib/zip_source_file_win32.h
@@ -59,6 +59,7 @@ struct zip_win32_file_operations {
     BOOL(__stdcall *move_file)(const void *from, const void *to, DWORD flags);
     BOOL(__stdcall *set_file_attributes)(const void *name, DWORD attributes);
     char *(*string_duplicate)(const char *string);
+    HANDLE(__stdcall *find_first_file)(const void *name, void *data);
 };
 
 typedef struct zip_win32_file_operations zip_win32_file_operations_t;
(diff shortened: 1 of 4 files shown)


ktakahashimtb:
I found that a similar fix is applied to other open source projects, and their commit description goes into very details on how a file on data-dedup-enabled filesystem becomes a reparse point, but the file should be regarded as a regular file.

* boost/filesytem: https://github.com/boostorg/filesystem/commit/141727b568fad2fecb77b2233a407cf4c00638b8
* golang: https://github.com/golang/go/commit/f26fa055229502fb6df4b6d9c300eccf01098248

Issue #451, opened 2024-08-22
`zip_source_filep_create` fails on Windows with a large zip archive


ktakahashimtb:
**Describe the Bug**
`zip_source_filep_create` fails on Windows when the size of zip archive is greater than ~2.14GB (== INT_MAX).

**Expected Behavior**
`zip_source_filep_create` and subsequent `zip_open_from_source` should be able to handle 2.14GB+ size archive on Windows.
The same archive can be opened via `zip_open` on Windows.

**Observed Behavior**
`zip_source_filep_create` fails with ZIP_ER_READ.

**To Reproduce**
On Windows, execute `zip_source_filep_create` with the file descriptor obtained for the large file.

**libzip Version**
v1.10.1 and also the tip of main

**Operating System**
Windows

**Test Files**
Any archive with size > 2.14GB.

**Additional context**
I'll follow up with a PR with a suggestion on how to address this limitation.

dillof:
We just committed a change that should fix this. Could you please test it in your setup.

PR #452, opened 2024-08-22
Fix file size limitation with `zip_source_filep_create` on Windows.

ktakahashimtb:
Proposed fix for #451 

On Windows, we need to call a version of `stat`, `fstat` and `fseek` which use 64 bit integer for file size and offset. Those being invoked currently use `long` type, which is 32 bit integer on both 32 and 64-bit Windows.


0-wiz-0:
We addressed this differently, see #451

PR #486, opened 2025-02-22
FindNettle: Prefer pkg-config Nettle version

cjmayo:
Gentoo Linux rewrites /usr/include/nettle/version.h to support multiple ABIs.

Even with Nettle 3.10 installed this results in:
-- Could NOT find Nettle: Found unsuitable version ".", but required is at least "3.0" (found /usr/lib64/libnettle.so)

---

I tried to otherwise maintain the logic; there might be other choices depending on when `set(Nettle_VERSION "1.0")` is wanted.


dillof:
Thanks, looks good, I've merged it.

PR #494, opened 2025-05-28
Use MbedTLS 3.6

winterheart:
Older versions of MbedTLS reached EOL and not supported anymore. MbedTLS 3.6 provides own configuration module for CMake, but some distributions may not iclude it into mbedtls package, so it is reasonable to keep own FindMbedTLS for compatibility.

Changed MbedTLS::MbedTLS target to MbedTLS::mbedcrypto to follow upstream naming.


0-wiz-0:
I'm fine with adapting the names to ones used by mbedtls upstream, but I don't see why we should require 3.6 now. The code uses APIs that were available since 1.0, so that's what we should require.
(If someone should really install such an old version of mbedtls is a different question.)

PR #498, opened 2025-06-26
Add OHOS systemName  matches to the CMakeLists.txt for Cross-Compilemethod of OpenHarmony

SC404:
we are using libzip in OpenHarmony, so we'd like to contribute the build guide, hope it can help developers.


0-wiz-0:
Merged, thank you!

PR #501, opened 2025-09-03
Address remaining issue with #451

ktakahashimtb:
This PR updates the fix for #451, where `zip_source_filep_create` fails on Windows with a large zip archive.

1.  Assign `_fseeki64` to `zip_os_fseek`.  `zip_os_seek` is not used anywhere in the code and `HAVE__SEEK64` is not defined  so assuming this was just an error.
2. `_ftelli64` needs to be assigned to `zip_os_ftell` so 64 bit version is invoked.


ktakahashimtb:
> The basic change looks fine, just one question: why do you move `#include <sys/stat.h>` to also apply on Windows?

Yes, _stat64 is defined in this header under Windows, too.  Without this include, `zip_source_file_stdio.c` does not compile, emitting "struct '_stat64' undefined" error where `zip_os_stat_t` type is used.
 
[Here is an example from Microsoft](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=msvc-170#example), where this header is used under Windows.

ktakahashimtb:
@dillof Thank you!

PR #506, opened 2025-09-30
Fix memory leak on multiple call to zip_file_set_encryption

remicollet:
If password is set twice, the first allocation is never freed

Reported as https://github.com/php/php-src/issues/19932

PHP use as workaround to free the first password:

     zip_file_set_encryption(intern, index, ZIP_EM_NONE, NULL)


dillof:
Thanks, good catch.

PR #520, opened 2026-03-15
Validate PKWARE compressed size before header subtraction

rootvector2:
PKWARE encrypted entries include a 12-byte encryption header that is
accounted for by subtracting `ZIP_CRYPTO_PKWARE_HEADERLEN` from
`st->comp_size` in the `ZIP_SOURCE_STAT` handler.

If a malformed archive reports a compressed size smaller than the
PKWARE header length, the subtraction may wrap because `comp_size`
is an unsigned integer.

This patch adds validation during source creation to ensure that the
compressed size is present and large enough to contain the PKWARE
header. A guard is also added in the STAT handler as a defense-in-depth
measure.

The change mirrors the validation pattern already used in
`zip_source_winzip_aes_decode.c`.


dillof:
Thank you, we've added your second check to winzip aes as well.

PR #521, opened 2026-03-18
cmake: bump minimum required for examples to version 3.5.0

florianernst:
This follows #392, and it was originally reported to the [Debian BTS](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1125640) but somehow not forwarded until now.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,3 +1,4 @@
+cmake_minimum_required(VERSION 3.5)
 foreach(PROGRAM add-compressed-data autoclose-archive in-memory)
     add_executable(${PROGRAM} ${PROGRAM}.c)
     target_link_libraries(${PROGRAM} zip)


0-wiz-0:
Can you please explain why this is needed?
The main CMakeLists.txt file already has
```
cmake_minimum_required(VERSION 3.10)
```

florianernst:
> Can you please explain why this is needed? The main CMakeLists.txt file already has [...]

Specifying `cmake_minimum_required` there is needed when not building the full library but instead only building the examples, e.g. as part of some CI, as e.g. done in the Debian project:

There, whenever any of libzip's (build-)dependencies change, as part of the so-called [autopkgtest CI](https://wiki.debian.org/autopkgtest) the libzip examples will be built and (partially) executed following [debian/tests/upstream](https://sources.debian.org/src/libzip/1.11.4-2/debian/tests/upstream), cf. [this job](https://ci.debian.net/packages/libz/libzip/unstable/amd64/69522497/) for one such execution (unfortunately, browsing the full data is only available to authenticated users after bots kept hammering the systems).
This happens in addition to regular full rebuilds of all software packages contained in Debian, in order to catch any regressions early. And during such a full rebuild of libzip the extra `cmake_minimum_required` doesn't hurt either, even though it is indeed not required.

And concerning the actual version for `cmake_minimum_required` you are right, the main CMakeLists.txt file already specifies a newer version. I simply missed to check that when forwarding the patch after seeing 3.5 also referenced in [examples/cmake-project/CMakeLists.txt](https://github.com/nih-at/libzip/blob/main/examples/cmake-project/CMakeLists.txt#L1), but it makes sense to keep them in line. I can adjust the patch, or you could simply adjust to your liking, considering that such an one-liner doesn't really require attribution.

0-wiz-0:
Thanks for the explanation. I've added the line.

PR #523, opened 2026-05-04
Fix zstd::libzstd_shared not found error on CYGWIN

carlo-bramini:
I tried to update the package of libzip for CYGWIN, but I got an error when I executed ninja for building.
When CMake configures the build, it prints this message on the console:
```
-- Found zstd: /usr/lib/libzstd.dll.a (Required is at least version "1.4.0")
```
and I have installed version zstd-1.5.7 in my environment.
However, when I start the real build process, it prints this error message:
```
ninja: error: 'zstd::libzstd_shared-NOTFOUND', needed by 'lib/cygzip-5.dll', missing and no known rule to make it
```
The bug happens because CYGWIN uses DLLs and import libraries like MinGW.
So I fixed the bug into `cmake/Findzstd.cmake` with this patch.


0-wiz-0:
Thanks, merged.

PR #526, opened 2026-05-26
ossfuzz: add zip_read_metadata_fuzzer for extra-field and metadata paths

XananasX7:
## Summary

The existing OSS-Fuzz harnesses read file *content* via `zip_fopen_index()`/`zip_fread()`. They do not exercise the metadata accessor APIs that parse ZIP extra fields in the central directory and local file headers.

## Coverage Gap

The following APIs are **not reached** by any existing harness:
- `zip_stat_index()` — parses stat info from central directory entries
- `zip_file_get_external_attributes()` — reads OS/attribute bytes
- `zip_get_archive_comment()` / `zip_file_get_comment()`
- `zip_file_extra_field_get()` — extra field by index (local + central)
- `zip_file_extra_field_get_by_id()` — by ID: Zip64 (0x0001), UNIX (0x000d), extended timestamp (0x5455), new Unix (0x7875), NTFS (0x000a), AES (0x9901)
- `zip_file_extra_fields_count()` for both `ZIP_FL_LOCAL` and `ZIP_FL_CENTRAL`

These paths live in `zip_extra_field.c` and `zip_dirent.c` and involve separate length/bounds arithmetic that is independent of the decompression path.

## New Harness: `zip_read_metadata_fuzzer.c`

Opens the fuzz input as a read-only archive, then for each entry exercises all the above accessors. Also reads the archive-level comment. Included in `ossfuzz/CMakeLists.txt` so it is automatically built and deployed by OSS-Fuzz.

## Testing

Built with `clang -fsanitize=address,undefined,fuzzer` against a static libzip built from source.

- ✅ No crashes found after 5 minutes / ~150,000 executions
- ✅ 292 corpus entries discovered; 2005 edge coverage
- ✅ Builds cleanly, one sign mismatch warning fixed

## Reward Program

Submitted under Google's [Patch Rewards Program](https://bughunters.google.com/patch-reward) (libzip is a Tier 2 OSS-Fuzz project).

### DIFF
--- a/ossfuzz/CMakeLists.txt
+++ b/ossfuzz/CMakeLists.txt
@@ -1,6 +1,7 @@
 set(FUZZ_PROGRAMS
   zip_read_file_fuzzer
   zip_read_fuzzer
+  zip_read_metadata_fuzzer
   zip_write_encrypt_aes256_file_fuzzer
   zip_write_encrypt_pkware_file_fuzzer
 )
--- a/ossfuzz/zip_read_metadata_fuzzer.c
+++ b/ossfuzz/zip_read_metadata_fuzzer.c
@@ -1,0 +1,129 @@
+/*
+  zip_read_metadata_fuzzer.c -- fuzz metadata parsing paths in libzip
+  Copyright (C) 2026 Google LLC
+
+  SPDX-License-Identifier: BSD-3-Clause
+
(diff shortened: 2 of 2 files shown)


dillof:
Thank you, looks good, and more coverage is always welcome.

PR #527, opened 2026-05-28
zero pkware password before free in trad_pkware_free

rootvector2:
winzip_aes_free and _zip_dirent_finalize zero the strdup'd password
with _zip_crypto_clear before free; the two trad_pkware_free helpers
in zip_source_pkware_(de|en)code drop the password directly, leaving
the plaintext in heap memory after free. Apply the same clear-then-free
pattern in both pkware paths so all four encryption contexts match.


dillof:
Good catch, thanks.

PR #528, opened 2026-05-29
clear derived key buffer in _zip_winzip_aes_new

rootvector2:
The PBKDF2 output buffer in zip_winzip_aes.c holds the derived AES and HMAC keys but isn't wiped before _zip_winzip_aes_new returns. Clear it with _zip_crypto_clear on each exit path, matching the ctx clear already done a few lines below.


dillof:
Thanks again.

PR #529, opened 2026-05-29
fix out-of-bounds read in zip_dir_add for empty name

rootvector2:
zip_dir_add() reads name[len - 1] to test for a trailing slash, but len is strlen(name) and is never checked for zero. Passing an empty name makes this name[(size_t)-1], a one-byte out-of-bounds read before the buffer. Guard the access with len == 0 so empty names take the trailing-slash path instead.


dillof:
Another good catch, thanks.

PR #531, opened 2026-06-01
Guard ZIP_CHECKCONS Overflow

metsw24-max:
Add overflow checks in ZIP_CHECKCONS validation paths to prevent malformed Zip64 archives from bypassing consistency checks through unsigned integer wraparound.

## Problem

`_zip_checkcons()` computes the end position of each entry using values read from the central directory:

```c
offset + comp_size + filename_length + LENTRYSIZE
```

For Zip64 archives, `comp_size` can be a full 64-bit value. A crafted archive can cause this calculation to overflow `zip_uint64_t`, wrapping to a small value and allowing an entry that actually extends beyond the central directory to pass ZIP_CHECKCONS validation.

A similar unchecked addition exists in `_zip_read_eocd64()` when validating the relationship between the EOCD64 record and EOCD64 locator.

### `lib/zip_open.c`

* Add overflow guards before calculating an entry's end position in `_zip_checkcons()`.
* Add overflow guards before evaluating `size + eocd_offset + 12` in `_zip_read_eocd64()`.
* Return an error when overflow is detected instead of allowing wrapped values to participate in consistency validation.
* Document the rationale with inline comments.

### Regression Test

Added:

* `regress/incons-zip64-checkcons-overflow.zip`
* `regress/open_incons_zip64_overflow.test`

The fixture contains a minimal Zip64 entry whose compressed size is supplied through a ZIP64 extra field with a value large enough to trigger the overflow condition. The regression test verifies that the archive is rejected when opened with `ZIP_CHECKCONS`.

## Impact

This change hardens ZIP_CHECKCONS against integer-overflow-based validation bypasses. The issue affects archive consistency verification and does not result in memory corruption or code execution, but it can cause malformed archives to be incorrectly accepted during consistency checks.


dillof:
Thank you, especially for also adding a test case.

We've modified it slightly to fit our style.

PR #532, opened 2026-06-01
zero default password before free

rootvector2:
za->default_password keeps the strdup'd cleartext decryption password (set via zip_set_default_password, used in zip_close.c and zip_fopen.c) but is freed without wiping in zip_set_default_password and zip_discard, unlike every other password/key path which already calls _zip_crypto_clear before free.


dillof:
Thanks. We've adapted it to use our idiom of NULL pointer check.

PR #533, opened 2026-06-01
don't read uninitialized window end for unbounded source

rootvector2:
A window source created with len -1 (end_valid false) never sets struct window::end, but window_read() reads it on ZIP_SOURCE_OPEN (ctx->end + offset) and on the SEEK_SET/SEEK_CUR fallthrough, where ctx->end - ctx->start is passed as the seek bound. Reachable through zip_source_window_create(src, start, -1) followed by zip_source_seek: the bound check then depends on uninitialized heap, so an in-range seek can be rejected or an out-of-range one accepted. Guard the open-time adjustment with end_valid and compute unbounded SEEK_SET/SEEK_CUR without referencing end.


dillof:
You're absolutely right that we shouldn't use end if it's invalid. However, we've implemented the fix by using the underlying source to validate seeks and using zip_source_seek_compute_offset when end is known.

PR #536, opened 2026-06-04
ossfuzz: add zip_write_roundtrip_fuzzer for the write/serialize paths

metsw24-max:
Add a new OSS-Fuzz target that exercises libzip's archive creation and serialization paths by performing a write → serialize → reopen → read round trip on fuzzer-provided input.

The existing write fuzzers (`zip_write_encrypt_aes256_file_fuzzer`, `zip_write_encrypt_pkware_file_fuzzer`) only add a single, fixed-metadata entry with one encryption method and never reopen the result. This target broadens write-side coverage to multi-entry archives, directories, deletes/renames, variable filenames, comments, extra fields, and compression/encryption method selection — then reads the generated archive back, exercising the encode → decode round trip.

## Changes

* Add `ossfuzz/zip_write_roundtrip_fuzzer.c`
  * Builds a ZIP archive from fuzzer input via a sequence of input-derived
    operations (add file/dir, set compression/encryption, set comments, set
    extra fields, rename, delete)
  * Serializes the archive with `zip_close()`
  * Reopens the resulting buffer and reads every entry back
* Register the target in `ossfuzz/CMakeLists.txt`

## Validation

* Builds with the OSS-Fuzz configuration (`BUILD_OSSFUZZ=ON`) and integrates
  with `fuzz_main.c`
* Standalone replay execution succeeds
* Ran ~450K executions under libFuzzer + AddressSanitizer with no crashes
  attributable to libzip

## Coverage

Locally, the target reaches roughly **2.3× the edge coverage** of the read-oriented harnesses (~9,300 vs ~4,050 features), entering archive creation, encoding, serialization, and round-trip code paths not covered by the read-focused fuzzers and only minimally touched by the existing single-entry encryption write fuzzers.

### DIFF
--- a/ossfuzz/CMakeLists.txt
+++ b/ossfuzz/CMakeLists.txt
@@ -4,6 +4,7 @@ set(FUZZ_PROGRAMS
   zip_read_metadata_fuzzer
   zip_write_encrypt_aes256_file_fuzzer
   zip_write_encrypt_pkware_file_fuzzer
+  zip_write_roundtrip_fuzzer
 )
 
 foreach(PROGRAM IN LISTS FUZZ_PROGRAMS)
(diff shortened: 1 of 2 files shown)


dillof:
Thank you, looks great. I especially like the idea of using the fuzzing data as commands.

PR #537, opened 2026-06-04
reject filenames longer than 65535 bytes in _zip_set_name

rootvector2:
noticed the `TODO: check for string too long` in `_zip_set_name` is real: it narrows `strlen(name)` to `zip_uint16_t`, so `zip_file_add`/`zip_dir_add` silently truncate a name over 65535 bytes to a different (possibly colliding) entry, while `zip_file_rename` already rejects it at its entry point, so add the same guard in the callee.


dillof:
Thanks for the fix, we rewrote it slightly to avoid calling strlen twice.

PR #538, opened 2026-06-04
Pass open flags to Zip64 EOCD consistency validation

jmestwa-coder:
## Summary

This change passes `za->open_flags` to `_zip_read_eocd64()` instead of `za->flags`.

## Problem

When a ZIP archive is opened with `ZIP_CHECKCONS`, `_zip_read_eocd64()` receives the internal archive flags rather than the user-specified open flags.

As a result, Zip64 EOCD consistency checks guarded by `ZIP_CHECKCONS` are not executed even when consistency checking is explicitly requested.

## Fix

* Pass `za->open_flags` to `_zip_read_eocd64()`.
* Ensure Zip64 EOCD consistency validation honors the flags provided by the caller.

## Testing

* Added a regression test covering a malformed Zip64 archive with inconsistent EOCD64 metadata.
* Verified the malformed archive is rejected when opened with `ZIP_CHECKCONS`.
* Verified existing consistency-check regression tests continue to pass unchanged.


dillof:
Good catch. Thank you for the bug fix and test case.

PR #540, opened 2026-06-13
ga: enable windows and fix build

scribam:


0-wiz-0:
Thank you! That did indeed fix the build, and we have merged the changes manually.

PR #541, opened 2026-06-15
Avoid Win32 file source I/O length truncation

Alb3e3:
## Summary

- cap Win32 file-source reads to the maximum `DWORD` byte count accepted by `ReadFile`
- write Win32 file-source output in `DWORD`-sized chunks instead of casting the full `zip_uint64_t` length
- preserve the existing write-all-or-fail behavior while avoiding large-length wraparound/truncation

## Rationale

The libzip source API uses `zip_uint64_t` byte counts, but the Win32 `ReadFile` and `WriteFile` APIs accept `DWORD` byte counts. Passing a larger value through a direct cast can wrap or truncate the transfer size before it reaches the OS.

For reads, returning a bounded chunk lets `zip_source_read()` continue its existing loop. For writes, the Win32 backend now loops over bounded chunks so large writes either complete the requested length or fail with the underlying write error.

## Testing

- `git diff --check`

Not run locally: full CMake/tests or Windows compile, because this container does not have `cmake`, `clang-format`, `cppcheck`, or a MinGW/Windows compiler installed.

### DIFF
--- a/lib/zip_source_file_win32.c
+++ b/lib/zip_source_file_win32.c
@@ -84,8 +84,7 @@ void _zip_win32_op_close(zip_source_file_context_t *ctx) {
 zip_int64_t _zip_win32_op_read(zip_source_file_context_t *ctx, void *buf, zip_uint64_t len) {
     DWORD i;
 
-    /* TODO: cap len to "DWORD_MAX" */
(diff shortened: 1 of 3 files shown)


dillof:
Thanks, we've implemented it slightly differently.

PR #546, opened 2026-06-26
run check_eocd consistency check under ZIP_CHECKCONS

arshsmith1:
check_eocd validates that the central directory ends exactly where the EOCD record begins, but only when ZIP_CHECKCONS is requested. It was passed za->flags instead of za->open_flags; bit 0x4 means ZIP_AFL_IS_TORRENTZIP in za->flags and ZIP_CHECKCONS in za->open_flags, and the torrentzip bit isn't set yet when the central directory is read, so the check never ran and an archive with a gap between the central directory and the EOCD was accepted even under ZIP_CHECKCONS. Pass za->open_flags here, the same way _zip_read_eocd64 was already corrected. The check also has to be skipped for Zip64 archives, where the central directory is followed by the Zip64 EOCD record and locator, so cd->offset + cd->size points before the 32-bit EOCD and the comparison would reject valid files.

### DIFF
--- a/lib/zip_open.c
+++ b/lib/zip_open.c
@@ -342,7 +342,7 @@ static bool _zip_read_cdir(zip_t *za, zip_buffer_t *buffer, zip_uint64_t buf_off
 
     /* We accept this EOCD as valid and won't search for an earlier one if it is unusable. */
 
-    if (!check_eocd(cd, za->flags, error)) {
+    if (!check_eocd(cd, za->open_flags, error)) {
         _zip_cdir_free(cd);
         return true;
     }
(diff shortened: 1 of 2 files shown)


arshsmith1:
Rebased onto current main to clear the AppVeyor failure. It was the `DWORD_MAX` undeclared error in zip_source_file_win32_named.c, not from this change: the branch was sitting on fdb9512, and 552e61f adds the intsafe.h include that fixes it. The patch itself is unchanged. Windows x64 builds clean and the full regress suite passes locally (186/186).

dillof:
Good catch, thanks.

We've improved your fix slightly and also fixed the wrong error code set in that case.

PR #548, opened 2026-07-10
check for overflow when computing seek offsets

arshsmith1:
zip_source_seek_compute_offset() forms the target as (zip_int64_t)offset + args->offset for SEEK_CUR and (zip_int64_t)length + args->offset for SEEK_END without guarding the addition, so a seek with a large delta (reachable via zip_fseek/zip_source_seek on a stored entry) overflows zip_int64_t; UBSan flags it at zip_source_seek.c:71. The window source repeats the same mistake in its SEEK_CUR guard, which does the overflowing add to detect the overflow. This checks the operands against ZIP_INT64_MAX/ZIP_INT64_MIN first, so out-of-range seeks return ZIP_ER_INVAL as before and valid seeks are unaffected.

### DIFF
--- a/lib/zip_source_seek.c
+++ b/lib/zip_source_seek.c
@@ -68,10 +68,18 @@ zip_int64_t zip_source_seek_compute_offset(zip_uint64_t offset, zip_uint64_t len
 
     switch (args->whence) {
     case SEEK_CUR:
+        if ((args->offset > 0 && (zip_int64_t)offset > ZIP_INT64_MAX - args->offset) || (args->offset < 0 && (zip_int64_t)offset < ZIP_INT64_MIN - args->offset)) {
+            zip_error_set(error, ZIP_ER_INVAL, 0);
+            return -1;
+        }
         new_offset = (zip_int64_t)offset + args->offset;
         break;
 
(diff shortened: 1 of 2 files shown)


arshsmith1:
any update?

dillof:
We didn't get to it yet. libzip is our hobby project, and we had a lot of pull requests lately.

dillof:
hi, we just committed a change based on PR #553, which should also fix the issues you reported. Could you please check if you see any remaining issues?

arshsmith1:
Checked the current main. Both spots I flagged now route through the new _zip_offset_add helper, so the SEEK_CUR/SEEK_END additions happen in unsigned space with the range checks done before the add. I rebuilt the static lib with -fsanitize=undefined and reran my two reproducers (the seek path in zip_source_seek and the unknown-length window SEEK_CUR case): both return ZIP_ER_INVAL now with no UBSan diagnostic, and valid set/cur/end seeks still land where they should. So no remaining issues on my end, your change resolves it. This PR is redundant now, feel free to close it out.

0-wiz-0:
Thanks for testing and the original bug report!

PR #549, opened 2026-07-13
Limit xz decoder memory usage

M0nd0R:
This change bounds XZ/LZMA decoder memory usage during reads and adds a regression for crafted archives with oversized dictionary headers.

The root cause is that `zip_algorithm_xz.c` initialized both decoders with `UINT64_MAX`, so attacker-controlled stream parameters could drive multi-gigabyte allocation attempts before libzip rejected the input. This patch caps decompressor memory to 128 MiB, which still leaves headroom above libzip's own encoder presets, and treats memlimit failures as compressed-data errors.

Impact:
- crafted XZ/LZMA members fail closed instead of attempting unbounded decoder allocations
- callers receive a normal compressed-data failure instead of resource blowup behavior
- a regression now covers an archive with an oversized dictionary header

Validation:
- `cmake --build build-pr --target lzma_large_dict -j4`
- `./build-pr/regress/lzma_large_dict`


dillof:
We understand your concern, but we don't want to reject valid files.

As a compromise, we've made this a compile time option, off by default.

PR #550, opened 2026-07-13
Fix zip_source_is_seekable for non-layered sources

M0nd0R:
This change fixes `zip_source_is_seekable()` for non-layered sources and adds a regression that covers both buffer-backed and root callback sources.

The root cause is that the public helper dereferenced `src->src` unconditionally. That only works for layered sources; root sources have no lower layer and can trigger a null dereference. This patch queries `zip_source_supports(src)` directly so the API works for both layered and non-layered sources.

Impact:
- `zip_source_is_seekable()` no longer crashes on root sources
- callers receive the expected seekability result for custom callback sources
- the regression covers both seekable and non-seekable top-level sources

Validation:
- `cmake --build build-pr --target source_is_seekable -j4`
- `./build-pr/regress/source_is_seekable`

### DIFF
--- a/lib/zip_source_supports.c
+++ b/lib/zip_source_supports.c
@@ -67,5 +67,5 @@ ZIP_EXTERN zip_int64_t zip_source_make_command_bitmap(zip_source_cmd_t cmd0, ...
 
 
 ZIP_EXTERN int zip_source_is_seekable(zip_source_t *src) {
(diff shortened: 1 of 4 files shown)


dillof:
Thanks for the bug report, we've applied your fix.

We don't think it's worth adding an extra regression program for this.

PR #551, opened 2026-07-13
Reject local header mismatches on read

M0nd0R:
This change rejects local/central header mismatches during ordinary reads and adds a regression that exercises a crafted archive with a forged local-header pointer.

The root cause is that `_zip_file_get_offset()` trusted the local header at the central-directory offset without verifying that the parsed local metadata still matched the central entry. `ZIP_CHECKCONS` already rejected this archive shape, but default reads could still follow a mismatched local header and open the wrong payload. This patch parses the local header, compares the relevant metadata against the central entry, and returns an inconsistency error when they diverge.

Impact:
- default reads now fail closed on forged local-header redirections
- behavior is aligned with `ZIP_CHECKCONS` for this inconsistency class
- the regression creates and patches a synthetic archive that previously opened the wrong file data

Validation:
- `cmake --build build-pr --target local_header_mismatch -j4`
- `./build-pr/regress/local_header_mismatch`

### DIFF
--- a/lib/zip_file_get_offset.c
+++ b/lib/zip_file_get_offset.c
@@ -37,6 +37,31 @@
 
 #include "zipint.h"
 
+static bool
+local_header_matches_central(const zip_dirent_t *central, const zip_dirent_t *local) {
+    if ((central->version_needed < local->version_needed)
+        || (central->comp_method != local->comp_method)
+        || (central->last_mod.time != local->last_mod.time)
+        || (central->last_mod.date != local->last_mod.date)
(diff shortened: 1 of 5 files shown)


dillof:
libzip's default behavior is to be permissible in what slightly malformed zip archives it accepts. There are a wide variety of programs creating zip archives, and not all of them produce perfectly correct archives. Precluding users of our library from working with those reduces the versatility of libzip.

If you want to be really picky and only accept perfect archives, use `ZIP_CHECK_CONS`, but don't force it on everyone.

PR #552, opened 2026-07-13
Report deferred read integrity errors on close

M0nd0R:
This change makes `zip_fclose()` report deferred integrity failures when a caller reads exactly the advertised file size, and adds a regression that covers CRC and encrypted-HMAC failures.

The root cause is that integrity errors can remain pending until the underlying source is asked for one more byte at end-of-stream. If the caller stops after an exact-size `zip_fread()`, libzip currently frees the source without surfacing the deferred failure. This patch performs a final probe in `zip_fclose()` when no prior error is recorded and the source position matches the advertised size.

Impact:
- exact-size reads no longer mask deferred CRC/HMAC/length failures
- callers now see the same integrity result whether they over-read by one byte or stop at the reported size
- regressions cover stored, deflated, and encrypted cases

Validation:
- `cmake --build build-pr --target exact_read_integrity -j4`
- `cd regress/data && ../../build-pr/regress/exact_read_integrity`

### DIFF
--- a/lib/zip_fclose.c
+++ b/lib/zip_fclose.c
@@ -36,6 +36,30 @@
 
 #include "zipint.h"
 
+static void
+check_for_pending_read_error(zip_file_t *zf) {
+    zip_stat_t st;
+    zip_int64_t position;
+    zip_int64_t n;
+    zip_uint8_t dummy;
+
+    zip_stat_init(&st);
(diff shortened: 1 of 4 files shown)


dillof:
That's a good catch, thank you.

However, we've implemented this at the `zip_source_t` layer, so it also works if you use sources directly instead of using `zip_fopen()`.

PR #553, opened 2026-07-13
Fix large-offset source seek arithmetic

M0nd0R:
## Summary
- make source seek offset arithmetic overflow-safe for large logical offsets
- fix unknown-length window sources so `SEEK_SET` and `SEEK_CUR` stay relative to the window start
- add a regression program covering unknown-length window seeks and large fragment-backed sources

## Verification
- `cmake -S . -B build -DBUILD_REGRESS=ON -DBUILD_OSSFUZZ=OFF -DBUILD_DOC=OFF -DBUILD_EXAMPLES=OFF`
- `cmake --build build --target zip -j4`
- `cc -I$(pwd)/lib -I$(pwd)/build -L$(pwd)/build/lib -Wl,-rpath,$(pwd)/build/lib -o /tmp/source_seek regress/programs/source_seek.c -lzip`
- `/tmp/source_seek ignored`

## Notes
- `nihtest` is not available in this environment, so CMake disables the full regress target here.

### DIFF
--- a/lib/zip_source_seek.c
+++ b/lib/zip_source_seek.c
@@ -35,6 +35,29 @@
 #include "zipint.h"
 
 
+bool _zip_offset_add(zip_uint64_t offset, zip_int64_t delta, zip_uint64_t *result) {
+    if (delta >= 0) {
+        zip_uint64_t unsigned_delta = (zip_uint64_t)delta;
+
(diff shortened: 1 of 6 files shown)


dillof:
Thanks. We've adapted your solution, and further simplified seek handling in zip_source_window.

PR #555, opened 2026-07-21
Fix source error handling across reopen

carrerasdarren-cell:
Follow-up to #552 and bf51731.

This PR originally covered two reopen regressions. Upstream commit 4cd7310 has now adopted the lookahead reset from this PR; this refresh rebases on current `main` and keeps regression coverage for it.

The remaining functional changes are:

- `zip_source_close()` no longer queries `ZIP_SOURCE_ERROR` after a callback returned success. Callback failures are already captured when they return `-1`; querying after success can resurrect an error from an earlier open cycle.
- The WinZip AES source explicitly reports deferred integrity failures from `ZIP_SOURCE_CLOSE`, clears its private error on reopen, and releases its crypto context on close. These changes preserve HMAC error propagation after removing the unconditional error query and avoid leaking the prior context across a partial-read reopen.

Regression coverage includes the lookahead case fixed by 4cd7310 and the still-distinct stale-error case. On current unmodified `main` plus the regression test, the latter fails with `closing reopened source failed: Invalid argument`; this patch passes it.

Validation:

- `ctest --test-dir build -R "^exact-read-integrity\\.test$" --output-on-failure`: passed.
- `ctest --test-dir build --parallel 4 --output-on-failure`: 187/187 passed.
- The prior AppVeyor build passed all x64 and Win32 jobs. Its two ARM jobs failed before project configuration because the runner lacks the selected Windows ARM SDK; current upstream `main` build 1.0.1028 fails in the same jobs for the same reason.

### DIFF
--- a/lib/zip_source_close.c
+++ b/lib/zip_source_close.c
@@ -56,13 +56,8 @@ int zip_source_close(zip_source_t *src) {
         }
     }
 
-    if (!_zip_source_had_error(src)) {
-        /* If the last read caused an error, get it. */
-        _zip_source_get_error(src);
-
-        if (ZIP_SOURCE_IS_LAYERED(src) && _zip_source_had_error(src->src)) {
-            zip_error_set_from_source(&src->error, src->src);
-        }
+    if (!_zip_source_had_error(src) && ZIP_SOURCE_IS_LAYERED(src) && _zip_source_had_error(src->src)) {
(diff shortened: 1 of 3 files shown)


dillof:
Thanks for the bug report and test program.

We found a solution that more closely resembles how errors are handled in other zip_source functions: Only propagate the error from the source if it returns -1.

Your extended tests pass with our change.

PR #557, opened 2026-07-28
Require Zip64 EF for 0xFFFFFFFF sentinels and check EF size overflow

M0nd0R:
## Summary
- Fail closed when a central/local header uses the `0xFFFFFFFF` Zip64 sentinel without a Zip64 extra field (APPNOTE requires the EF; without it sizes stay fake 4GiB and data-descriptor width is wrong).
- Check for unsigned overflow when summing extra-field lengths before applying the `ZIP_UINT16_MAX` cap (addresses the existing TODO in `_zip_dirent_write`).
- Validate `LOCAL_MAGIC` and that the local filename length matches the central directory before returning a data offset.

## Test plan
- [ ] Build + existing regress suite
- [ ] Archive with Zip64 sentinels but missing EF is rejected (`ZIP_ER_INCONS` / `INVALID_ZIP64_EF`)
- [ ] Valid Zip64 archives still open and extract
- [ ] Mismatched local vs central filename length is rejected on data access

### DIFF
--- a/lib/zip_dirent.c
+++ b/lib/zip_dirent.c
@@ -578,18 +578,26 @@ zip_int64_t _zip_dirent_read(zip_dirent_t *zde, zip_source_t *src, zip_buffer_t 
         zde->comment = utf8_string;
     }
 
-    /* Zip64 */
-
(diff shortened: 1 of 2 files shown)


dillof:
Thank you for the bug reports. 

However, please create separate pull requests for each issue. That makes it much easier for us to handle.

We've used different implementations to fix the errors.

PR #564, opened 2026-08-19
restore upper bound check in zip_source_seek_compute_offset

arshsmith1:
Commit 35c85c4 split zip_source_seek_compute_offset into per-whence cases but only kept the new_offset > length bound for SEEK_SET, so a forward SEEK_CUR or SEEK_END that stays under ZIP_INT64_MAX is accepted even when it lands past the end of the source. Reading the local extra fields of a crafted archive opened from an in-memory buffer source seeks by the local header's filename length, which then moves past buffer->size and buffer_read does an out-of-bounds heap read (ASan reports a heap-buffer-overflow in buffer_read). This restores the upper bound the pre-35c85c4 code applied to every whence.


dillof:
Sigh, thanks! 

I've committed the fix. The test case is too indirect for my taste. We should come up with a framework to test sources directly.

arshsmith1:
Thanks for committing this. Fair point on the test being indirect, a way to exercise sources directly would make cases like this much cleaner to cover.

PR #565, opened 2026-08-21
Reject ZIP_FL_LOCAL|ZIP_FL_CENTRAL in zip_file_extra_field_get_by_id()

Nexory:
Since 1.12 keeps local and central extra fields in separate lists, an extra
field index can no longer address both at once: the index is applied to each
list in turn and the first match wins, so fields behind that match are skipped.

With one local and two central fields of the same id:

```
count_extra_by_id  l / c / cl   ->  1 / 2 / 3
get_extra_by_id 0 cl            ->  local,      len 9
get_extra_by_id 1 cl            ->  central #1, len 16
get_extra_by_id 2 cl            ->  ZIP_ER_NOENT
```

The count reports three fields, but the first central one is not reachable at
any index, and index 1 returns a valid pointer to a different field rather than
an error. Addressing one list at a time still finds all three.

zip_file_extra_field_get() got a guard for exactly this in cbad2ba5. The by_id
variant was not changed, so this adds the same guard there. Addressing a single
list is unchanged.

I looked at the rest of the family before assuming it was wider: the two count
functions take no index, so both lists are unambiguous for them, and
zip_file_extra_field_set() with both flags writes the field to both lists,
which is lossless. Only this one function needed the guard.

The behaviour is on main only, so no released version is affected. I built with
BUILD_DOC=OFF because the documentation step writes generated files back into
the source tree, which I mount read-only.


dillof:
Good catch, but we decided to fix the API for ZIP_EF_BOTH instead. This avoids breaking existing code.

PR #566, opened 2026-08-22
Fix extra field setter return value

carrerasdarren-cell:
The extra-field refactor changed `zip_file_extra_field_set()` from returning `0` directly to returning the boolean result of `_zip_extra_fields_set()`. This inverted the public API contract: successful calls returned `1` instead of `0`, while failures returned `0` instead of `-1`. In particular, an invalid extra-field index was reported as success.

Map the internal boolean to the documented public return values and add a regression test for the invalid-index path. This behavior is on current main only, so no released version is affected.

Verification:
- `ctest --test-dir build-asan --output-on-failure -j4`
- 191/191 tests passed with AddressSanitizer and UndefinedBehaviorSanitizer enabled.

Assisted-by: OpenAI Codex

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # 1.12 [Unreleased]
 
+* Fix return value of `zip_file_extra_field_set()`.
 * Keep order of local extra fields, even if an identical extra field is present in the central directory.
 * Add `zip_source_at_eof()` and `ZIP_SOURCE_AT_EOF`.
 * Consistently report CRC and decryption errors when reading exactly to end-of-file.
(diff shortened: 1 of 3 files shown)


carrerasdarren-cell:
CI note: the prior AppVeyor run completed 6 of 8 matrix jobs successfully. The only failures were the ARM and ARM/UWP jobs, both during toolchain setup before libzip configured: the image lacks Windows SDK 10.0.26100.0 for Desktop C++ ARM Apps, and vcpkg consequently cannot configure bzip2/its compiler probe. Current `main` and unrelated PRs fail in the same two jobs, so this is an AppVeyor image issue rather than a patch failure. The patch itself passed the full local ASan/UBSan suite (191/191). I also cleaned the commit-message formatting without changing the tree.

dillof:
Good catch, thanks.

Since this introduced no change since the last release, it shouldn't be mentioned in NEWS.md

PR #567, opened 2026-08-22
Reject window stat past lower source end

carrerasdarren-cell:
## Summary

- reject an unknown-length window whose start exceeds the lower source size during `zip_source_stat()`
- prevent the unsigned subtraction from reporting `ZIP_UINT64_MAX` (or another wrapped size)
- add a public-API regression covering the out-of-range start

Before this change, a one-byte buffer wrapped with `zip_source_window_create(base, 2, -1, ...)` reported a size of `18446744073709551615`. File-backed sources already reject the equivalent range with `ZIP_ER_INVAL`; the window source now does the same when its lower source exposes a size.

## Verification

- built the library with CMake
- compiled and ran `regress/programs/source_seek.c` directly
- repeated the focused regression under AddressSanitizer and UndefinedBehaviorSanitizer

### DIFF
--- a/lib/zip_source_window.c
+++ b/lib/zip_source_window.c
@@ -325,6 +325,10 @@ static zip_int64_t window_read(zip_source_t *src, void *_ctx, void *data, zip_ui
                 st->size = ctx->end - ctx->start;
             }
             else if (st->valid & ZIP_STAT_SIZE) {
+                if (st->size < ctx->start) {
+                    zip_error_set(&ctx->error, ZIP_ER_INVAL, 0);
+                    return -1;
+                }
                 st->size -= ctx->start;
             }
         }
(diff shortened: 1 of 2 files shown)


carrerasdarren-cell:
CI note: AppVeyor build 1.0.1062 completed 6 of 8 matrix jobs successfully. The only failures were the ARM and ARM/UWP jobs, both during toolchain setup before libzip configured: the image lacks Windows SDK 10.0.26100.0 for Desktop C++ ARM Apps, and vcpkg consequently cannot configure bzip2/its compiler probe. Current `main` and unrelated PRs fail in the same two jobs, so this is an AppVeyor image issue rather than a patch failure. The focused source-window regression passes locally, including under ASan/UBSan.

PR #569, opened 2026-08-24
handle local zip64 extra field with 32-bit compressed size

arshsmith1:
A local Zip64 extra field always carries both the original and compressed size (appnote 4.5.3), but zip_dirent_process_ef_zip64 only consumes the compressed-size slot when the compressed size is the field that needs Zip64. For a large, highly compressible entry the uncompressed size overflows 32 bits while the compressed size still fits, so that 8-byte slot is left unread, the trailing length no longer matches got_len, and the entry is rejected as an invalid Zip64 extra field. It shows up when local headers are read, opening with ZIP_CHECKCONS.

Skip the compressed-size slot in the local case too, mirroring the branch just above that already skips the uncompressed-size slot when that one fits. Keeping both together keeps the "a local header must include both sizes" rule in one spot. The regress test is a valid archive whose local field holds both sizes with only the uncompressed size in Zip64 range; it fails before this change and passes after.


0-wiz-0:
Section 4.5.3 has two sentences that refer to this, this change request only seems to use one of the two.
Here are the two sentences I mean:
```
      The order of the fields in the zip64 extended
      information record is fixed, but the fields MUST
      only appear if the corresponding Local or Central
      directory record field is set to 0xFFFF or 0xFFFFFFFF.
```
and
```
      This entry in the Local header MUST include BOTH original
      and compressed file size fields.
```
I interpret this as: the local header must have both compressed and uncompressed sizes in the extra field.
So that this is allowed (per the first sentence), the record field must be 0xffffffff. The zip file you propose has the compressed size set to 20 - it should be 0xffffff instead.

arshsmith1:
Good point, that tension between the two sentences is exactly what this is about. The reason the test uses a compressed size of 20 rather than 0xffffffff is to mirror the branch right above the new one. The existing `else if (local)` skips the uncompressed slot when that size fits in 32 bits but the local field still carries it (sentence 2), and my change does the same for the compressed slot. So the archive has the uncompressed record at 0xffffffff and the compressed record at 20, which is the mirror image of the case that block already handles.

If I set the compressed record to 0xffffffff too, then the `comp_size == ZIP_UINT32_MAX` read branch consumes that slot and the new skip is never reached, so the test wouldn't actually exercise the fix. That archive already opens fine on current main.

If you'd rather lean on sentence 1 and require both record fields to be 0xffffffff whenever the local EF is present, then the existing uncompressed skip is in the same boat and we'd want to look at both together. I'm happy to go either way, just let me know which reading you prefer.

PR #571, opened 2026-08-29
Reset WinZip AES HMAC state when reopening a source

HyunsikParker:
Reset `hmac_verified` for every `ZIP_SOURCE_OPEN` cycle. Without the reset, reopening a WinZip AES source after an authentication failure skips HMAC verification on the second cycle because `hmac_verify_failed` is cleared while `hmac_verified` remains set.

The regression reuses existing AE-2 fixtures and checks that:

- corrupted empty and non-empty HMACs fail on both open cycles;
- a valid AES source succeeds on both open cycles.

Tests:

- normal suite: 190 tests, 0 failed;
- AddressSanitizer and UndefinedBehaviorSanitizer: 187 executed, 0 failed, 3 filesystem-clone tests skipped.


dillof:
Good catch, thanks.

PR #572, opened 2026-08-29
Updating functions to use bool where possible

brazosbuckaroo:
I made the functions I could use booleans. There were a couple functions that still needed to use int so I kept them as int. While going through this code, I noticed in zip_unchange_archive that it never really fails. It only ever returns 0. Was this purposeful or should I investigate it further? 

I want to ensure I am not going off in the weeds.

The function in question:
```
ZIP_EXTERN int zip_unchange_archive(zip_t *za) {
    if (za->comment_changed) {
        _zip_string_free(za->comment_changes);
        za->comment_changes = NULL;
        za->comment_changed = 0;
    }

    za->ch_flags = za->flags;

    return 0;
}
```

Lastly, I wlll start updating documentation if this pull request meet ya'lls standards.


dillof:
I'm sorry to say that this pull request is unacceptable for a couple of reasons:

1. Changing the argument or return types of public functions breaks backwards compatibility.
2. `compat.h` is a private include file that is not installed. It cannot be included in the public header `zip.h`.
3. `bool` is not available on all systems libzip supports (which is why we use `zip_bool` internally).
4. The debug `printf` in `zip_source_is_seekable` has already been removed on HEAD.

If you intend to make sweeping changes it would save us all time and effort if you asked us first in the Discussions section to see if it's something we actually want, and how it should be implemented.

brazosbuckaroo:
@dillof understood! apologies for that! It is my first time branching outside of my own code and tools. 

The only thing I will comment on is the printf statement. That was something I added for an unrelated thing and did not realize it had followed on that branch. I has curious about something. I removed it as soon as I realized. completely on me! So sorry!

PR #573, opened 2026-08-30
Reject zip source ranges starting past EOF

samuelgirmametaferia:
## Summary

Reject `zip_source_zip_file_create()` ranges whose starting offset is already beyond the entry size, including the `len == -1` case.

When the source size is known and `len == -1`, the existing validation does not check `start > st.size` before computing `st.size - start`. Because both values are unsigned, a sufficiently large start can wrap that subtraction to a small positive length and allow an invalid source window to be created.

The fix validates the starting offset independently of the requested length, before the subtraction. The regression uses the public API with `start == UINT64_MAX` on a small stored entry and requires `ZIP_ER_INVAL`.

## Validation

- CMake build: pass
- focused `source_seek` regression: pass
- current `origin/main` with the same regression: fails with `zip source allowed start past end of file`
- `git diff --check`: pass

PR #567 is related to out-of-range generic window sources, but fixes `zip_source_window.c` during `ZIP_SOURCE_STAT`; this change addresses the earlier range arithmetic in `zip_source_zip_file_create()` and does not overlap its code path.


samuelgirmametaferia:
Hey could you please review this!

dillof:
This is our hobby, not our job. We work on libzip in our free time, so it might take us a few weeks to get to all submissions.

samuelgirmametaferia:
Fair enough. I wasn’t trying to rush anyone, I was just asking for a review since the queue looked relatively small. I’ll leave it there and wait for someone to get to it.

dillof:
Thanks for the clarification. Your phrasing made it sound impatient. May I suggest something along the lines of "This PR is ready for review" next time?

We usually meet every 2 to 3 weeks to go through and discuss all PRs. Small straightforward fixes we sometimes merge in between, time allowing.

dillof:
Good catch, thanks.

We've implemented it slightly differently and used ziptool_regress for testing instead of a dedicated C program.

PR #575, opened 2026-09-03
fix: Reset CRC context state on source reopen

XananasX7:
## Summary

When a CRC source is reopened (`ZIP_SOURCE_OPEN`), only `position` was reset to zero. The remaining context fields — `crc_position`, `crc`, `crc_complete`, and `size` — retained stale values from the previous read pass.

## Problem

On a second open/read/close cycle, this could cause:

- CRC validation to be skipped (if `crc_complete` was still set from the prior pass)
- Incorrect size and CRC reported via `ZIP_SOURCE_STAT`
- Stale accumulated CRC used for validation of new data

This is the same class of bug as the WinZip AES HMAC state reset (commit e1f2793). That fix correctly reset `hmac_verified` on reopen, but the CRC source was not given the same treatment.

## Fix

Reset all CRC context fields on reopen to match the initial-state semantics expected by the layered source protocol.

## Testing

- Builds cleanly with ASAN/UBSAN (gcc 13, Ubuntu 24.04)
- The existing test suite passes


dillof:
Good catch, thanks. We've taken the opportunity to refactor the code to put all init/cleanup code into functions.

PR #576, opened 2026-09-06
Preserve explicit crypto buffer clearing across backends

agamal026:
## Summary

The build already probes for `explicit_memset` and `explicit_bzero`, but their results were missing from `config.h.in`. Propagate those results so the existing `_zip_crypto_clear` backend selection can use the available API.

For systems without either API, use volatile-qualified byte stores instead of ordinary `memset`. Keep `string.h` available independently of the selected backend, and use the `__inline` spelling for MSVC's default C mode. Existing call sites and the public API are unchanged.

Add `crypto_clear.test` for clearing an exact range, preserving adjacent guard bytes, and doing nothing for a zero-length range.

Also install nihtest with the Python 3.11 launcher matching AppVeyor's existing PATH, and fail the test-enabled jobs if CTest discovers no tests. The first Windows CI run compiled successfully but silently disabled regressions because unqualified `py` installed nihtest under Python 3.14.

## Validation

- Linux GCC 15.2 static-library builds with OpenSSL enabled, both native `explicit_bzero` and forced fallback.
- The new test passed through CMake/CTest and nihtest 1.11.1 in both configurations with `-O3 -flto -DNDEBUG`.
- Focused direct test builds with `-Wall -Wextra -Werror` passed in the preceding local validation.
- A separate GCC `-O3` assembly check retains all 16 clearing stores to an otherwise unused local buffer in the fallback.
- Windows Visual Studio 2019 x64 and x86 both executed and passed `crypto_clear.test` after the Python launcher correction.
- The CI configuration parses with only the two intended command changes. An empty CTest control returns success by default and failure with `--no-tests=error`.
- clang-format 19 and diff whitespace checks pass.

### Windows CI follow-up

[AppVeyor build 1.0.1075](https://ci.appveyor.com/project/nih-at/libzip/builds/54671467), for PR head `bea92fc25b3f2a371b1ac2ee8c53431e1fcc49ec`:

- x64 Windows: 191 registered tests, zero failures and five skips.
- x86 Windows: 191 registered tests, two failures and five skips. `crypto_clear.test` passed; the failures are `set_compression_store_to_xz.test` and `set_compression_xz_to_store.test`.
- ARM Windows and ARM UWP fail during toolchain setup, before libzip compilation, with `MSB8037` for Windows SDK `10.0.26100.0` Desktop C++ ARM Apps. The preceding build of this PR has the same setup failure; it is not an unmodified upstream baseline.

The x86 job uses liblzma 5.8.1. [XZ 5.8.2 release notes](https://github.com/tukaani-project/xz/blob/master/NEWS) describe a workaround for a compiler bug in 32-bit x86 CLMUL CRC builds with older MSVC. This is a possible explanation, not a confirmed cause of these failures; no corrected-dependency Windows run has verified it. No failing tests have been disabled.

This does not establish a passing full CI matrix or validate the `explicit_memset` backend. The change strengthens the existing clearing primitive; it does not guarantee erasure of every possible copy of sensitive data.

Implementation and review were AI-assisted.

### DIFF
--- a/lib/zipint.h
+++ b/lib/zipint.h
@@ -523,14 +523,29 @@ typedef struct _zip_pkware_keys zip_pkware_keys_t;
 #define ZIP_WANT_TORRENTZIP(za) ((za)->ch_flags & ZIP_AFL_WANT_TORRENTZIP)
 
 
+#include <string.h>
+
 #ifdef HAVE_EXPLICIT_MEMSET
 #define _zip_crypto_clear(b, l) explicit_memset((b), 0, (l))
 #else
 #ifdef HAVE_EXPLICIT_BZERO
 #define _zip_crypto_clear(b, l) explicit_bzero((b), (l))
 #else
-#include <string.h>
-#define _zip_crypto_clear(b, l) memset((b), 0, (l))
+#ifdef _MSC_VER
+static __inline
+#else
+static inline
+#endif
+    void
+    _zip_crypto_clear(void *buffer, size_t length) {
+    volatile unsigned char *p = (volatile unsigned char *)buffer;
+
+    /* Keep the clearing stores observable even when the buffer is no longer used. */
+    while (length > 0) {
+        *p++ = 0;
+        length--;
+    }
+}
 #endif
 #endif
 
--- a/regress/programs/crypto_clear.c
+++ b/regress/programs/crypto_clear.c
@@ -1,0 +1,56 @@
+/*
+  crypto_clear.c -- verify internal crypto buffer clearing
+  SPDX-License-Identifier: BSD-3-Clause
+
+  This file is part of libzip, a library to manipulate ZIP archives.
+*/
+
+#include "config.h"
+
+#include <stdio.h>
+
+#include "zipint.h"
+
+
+static int check_range(void) {
+    unsigned char buffer[] = {0xa5, 0x11, 0x22, 0x33, 0x44, 0x5a};
+    size_t i;
+
+    _zip_crypto_clear(buffer + 1, 4);
+
+    if (buffer[0] != 0xa5 || buffer[5] != 0x5a) {
+        fprintf(stderr, "crypto clear modified a guard byte\n");
+        return 1;
+    }
+    for (i = 1; i < 5; i++) {
+        if (buffer[i] != 0) {
+            fprintf(stderr, "crypto clear did not clear byte %zu\n", i);
+            return 1;
+        }
+    }
+
+    return 0;
+}
+
+
+static int check_zero_length(void) {
+    unsigned char buffer[] = {0xa5, 0x11, 0x22, 0x33, 0x44, 0x5a};
+    static const unsigned char expected[] = {0xa5, 0x11, 0x22, 0x33, 0x44, 0x5a};
+    size_t i;
+
+    _zip_crypto_clear(buffer + 2, 0);
+
+    for (i = 0; i < sizeof(buffer); i++) {
+        if (buffer[i] != expected[i]) {
+            fprintf(stderr, "zero-length crypto clear modified byte %zu\n", i);
+            return 1;
+        }
+    }
+
+    return 0;
+}
+
+
+int main(void) {
+    return check_range() || check_zero_length();
+}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -64,7 +64,7 @@ environment:
 
 before_build:
   cmd: >-
-    py -m pip install nihtest
+    py -3.11 -m pip install nihtest
 
     mkdir build
 
@@ -86,7 +86,7 @@ test_script:
   cmd: >-
     set VERBOSE=yes
 
-    IF %RUN_TESTS%==yes ( ctest -C %CMAKE_CONFIG% --output-on-failure )
+    IF %RUN_TESTS%==yes ( ctest -C %CMAKE_CONFIG% --output-on-failure --no-tests=error )
 
 cache:
   - c:\vcpkg.cache -> vcpkg.json
--- a/config.h.in
+++ b/config.h.in
@@ -27,6 +27,8 @@
 #cmakedefine HAVE_CLONEFILE
 #cmakedefine HAVE_COMMONCRYPTO
 #cmakedefine HAVE_CRYPTO
+#cmakedefine HAVE_EXPLICIT_BZERO
+#cmakedefine HAVE_EXPLICIT_MEMSET
 #cmakedefine HAVE_FICLONERANGE
 #cmakedefine HAVE_FILENO
 #cmakedefine HAVE_FCHMOD
--- a/regress/CMakeLists.txt
+++ b/regress/CMakeLists.txt
@@ -3,6 +3,7 @@ check_function_exists(getopt HAVE_GETOPT)
 set(TEST_PROGRAMS
   add_from_filep
   can_clone_file
+  crypto_clear
   exact_read_integrity
   fopen_unchanged
   fseek
--- a/regress/crypto_clear.test
+++ b/regress/crypto_clear.test
@@ -1,0 +1,3 @@
+description internal crypto clear zeroes only the requested range
+program crypto_clear
+return 0


regress/programs/crypto_clear.c
```diff
@@ -0,0 +1,56 @@
...
+            fprintf(stderr, "crypto clear did not clear byte %zu\n", i);
+            return 1;
+        }
+    }
+
+    return 0;
+}
+
+
+static int check_zero_length(void) {
```
dillof: Create one function that takes start and length as arguments. Copy expected into buffer, call _zip_crypto_clear() and check that the cleared range is 0 and the rest is unchanged.

lib/zipint.h
```diff
@@ -523,14 +523,29 @@ typedef struct _zip_pkware_keys zip_pkware_keys_t;
 #define ZIP_WANT_TORRENTZIP(za) ((za)->ch_flags & ZIP_AFL_WANT_TORRENTZIP)
 
 
+#include <string.h>
```
dillof: We don't want to expose <string.h> unless we have to.

lib/zipint.h
```diff
@@ -523,14 +523,29 @@ typedef struct _zip_pkware_keys zip_pkware_keys_t;
...
 #else
-#include <string.h>
-#define _zip_crypto_clear(b, l) memset((b), 0, (l))
+#ifdef _MSC_VER
+static __inline
+#else
+static inline
+#endif
+    void
+    _zip_crypto_clear(void *buffer, size_t length) {
```
dillof: Why does this function have to be inline?

I think volatile is needed so the compiler optimizer doesn't remove accesses to it. Why can't we call memset on the volatile pointer?

dillof:
Could you please look at my review comments?

Also, since you seem to know Appveyor, could you have a look at the two failing platforms? We would greatly appreciate any support on Windows.

agamal026:
Thanks for the follow-up. I checked the PR conversation, review threads, and inline comments, but I cannot see your review comments yet. Could you link them or repost them so I can address the specific points?

For Windows CI, [PR build 1.0.1075](https://ci.appveyor.com/project/nih-at/libzip/builds/54671467) has two failing x86 XZ tests; `crypto_clear.test` passes there. ARM and ARM-UWP stop before libzip compilation with `MSB8037` because the Desktop C++ ARM component of Windows SDK `10.0.26100.0` is unavailable. The same ARM setup error appears on the later [main build 1.0.1094](https://ci.appveyor.com/project/nih-at/libzip/builds/54748300), which points to the CI image/SDK rather than this patch, though I have not validated a fix. That main build's x86 job reported "No tests were found", so its green status is not a valid comparison for the two XZ failures. Their cause remains unconfirmed; I am investigating them separately.

dillof:
> Thanks for the follow-up. I checked the PR conversation, review threads, and inline comments, but I cannot see your review comments yet. Could you link them or repost them so I can address the specific points?

Sorry, my bad, I missed that I had to press the submit button at the top. They should be visible now.

Thanks for investigating the CI build issues, we greatly appreciate it.

regress/programs/crypto_clear.c
agamal026: Addressed in b321748. The single check_range(start, length) helper copies the expected bytes, clears the requested span, and compares every byte against zero inside or its original value outside. It covers an interior range, zero length, and the full buffer. The focused CTest passes in native static, forced-fallback static, and native shared builds.

lib/zipint.h
agamal026: Agreed, and addressed in b321748. zipint.h now declares the internal function without including <string.h>. The implementation includes it privately, and the source files that had relied on the transitive include now include the declarations they use directly. Local native and forced-fallback builds pass; Windows validation of this new head is pending.

lib/zipint.h
agamal026: Inline was only needed for the old header-local fallback; b321748 moves clearing into a compiled internal function. memset accepts non-volatile void *, so passing a volatile pointer discards that qualifier, and casting it away does not prevent a dead-buffer memset from being optimized out. The fallback therefore keeps volatile byte stores; explicit_memset and explicit_bzero remain preferred where available. The internal symbol stays hidden, and the focused test compiles the same implementation directly. Local static and shared tests pass; Windows CI for this head is pending.

dillof:
We've merged your `_zip_crypto_clear()` fixes, thanks.

Your Appveyor commits seem like they add debug output. Can you run them on your branch? Or do we need to enable that, or should we commit them on the main branch? Thanks for looking into it.

agamal026:
Yes. The temporary ARM failure diagnostics ran on this branch in [AppVeyor 1.0.1098](https://ci.appveyor.com/project/nih-at/libzip/builds/54763901). They showed vcpkg ARM test-link failures (kernel32.lib / WindowsApp.lib) before libzip compilation; they did not fix those jobs. I removed the diagnostic hook and unsuccessful ARM overrides in 3490217, so there is no debug output left to enable or merge. The current branch ran again in [1.0.1101](https://ci.appveyor.com/project/nih-at/libzip/builds/54764464): x64 passed 193 tests; x86 failed only the two XZ conversion tests; ARM and ARM-UWP still failed before libzip. The remaining AppVeyor test-discovery changes are already on main via b2da382. No further AppVeyor commit is needed from this branch.

agamal026:
Follow-up on the two x86 XZ failures: the [controlled comparison now passes](https://github.com/agamal026/libzip/actions/runs/35831846406), with the same libzip `b2da382`, MSVC 19.29.30159.0 and 32-bit objects in all three variants:

| liblzma build | Same five regression tests |
| --- | --- |
| 5.8.1, CLMUL CRC enabled | Only the two XZ conversions fail, with the same `3221225477` / `Compressed data invalid` signatures as AppVeyor |
| 5.8.1, CLMUL CRC disabled | 5/5 pass |
| 5.8.3, normal CLMUL option and upstream compiler workaround | 5/5 pass |

The five tests cover crypto clearing and both LZMA/XZ conversions; none are skipped. Source commit IDs, compiler metadata, generated test environments and JUnit output are retained in the run's `x86-xz-comparison` artifact. Two earlier harness setup failures remain documented on the diagnostic branch; they are not counted as comparison evidence.

This isolates the failing old-MSVC CLMUL path on this runner and matches [XZ 5.8.2's documented compiler workaround](https://github.com/tukaani-project/xz/releases/tag/v5.8.2). The [vcpkg revision already used by your GitHub workflow contains liblzma 5.8.3](https://github.com/microsoft/vcpkg/blob/f3e10653cc27d62a37a3763cd84b38bca07c6075/ports/liblzma/vcpkg.json), while AppVeyor 1.0.1108 used 5.8.1.

Updating AppVeyor's dependency snapshot is therefore a concrete fix to validate next. This comparison does not establish a full AppVeyor matrix fix or resolve the separate ARM SDK/linker failures. I have not changed the existing PR branch.

agamal026:
I pushed the AppVeyor dependency update and merged current main into the branch, keeping the integrated crypto implementation unchanged. The remaining PR diff is only seven added lines in `appveyor.yml`: use the same vcpkg snapshot as GitHub CI, which supplies liblzma 5.8.3.

The [full AppVeyor build 1.0.1111](https://ci.appveyor.com/project/nih-at/libzip/builds/54771511) has finished:

- x86 and x64 Windows each ran 193 CTest entries: 188 passed, five skipped, zero failed. Both XZ conversion tests and `crypto_clear.test` passed. The x86 log confirms MSVC 19.29.30159.0 and liblzma 5.8.3.
- x86-UWP, x64-UWP, ARM64 Windows and ARM64-UWP built successfully; those jobs do not run tests.
- The two ARM32 jobs still fail before libzip compilation and report `MSB8037` for the missing Windows SDK 10.0.26100.0 Desktop C++ ARM components. Those setup failures remain unresolved, so the overall build is still red.

This validates the dependency update against the actual x86 regressions; it is not a full ARM32 fix. No matrix entries or test conditions were removed or disabled.

agamal026:
The ARM32 SDK fix is now validated on the actual AppVeyor workers. [Build 1.0.1114](https://ci.appveyor.com/project/nih-at/libzip/builds/54772589), for head `5aeb57b`, completed successfully across all eight original matrix jobs.

- ARM Windows and ARM-UWP both initialize SDK 10.0.19041.0, complete the vcpkg dependency builds, and build libzip successfully. The logs confirm Desktop/UWP mode respectively and the same SDK selected by outer CMake.
- x86 and x64 Windows each ran 193 CTest entries: 188 passed, five skipped, zero failed. Both XZ conversion tests and `crypto_clear.test` passed.
- The other six rows are build-only, as before; their success does not represent runtime test coverage.

The remaining diff against current main is `appveyor.yml` plus four small ARM CI helper/triplet files. It retains the dependency snapshot update and selects the installed ARM SDK in both the outer build and vcpkg's compiler environment. Temporary diagnostics are removed; the integrated crypto implementation, eight matrix rows and test conditions are unchanged.

PR #577, opened 2026-09-10
reject NULL data with a positive length in zip_file_extra_field_set

seonglae:
`zip_file_extra_field_set` accepts `data == NULL` with `len > 0`. The three closest public siblings reject that pair: `zip_set_archive_comment`, `zip_file_set_comment` and `zip_source_buffer` all set `ZIP_ER_INVAL`.

Here the value reaches `_zip_extra_fields_set` and then `_zip_memdup`, which hands the unchecked pointer to `memcpy_s`, so with a length of 1 it is a one byte read from address zero. Reported by ASan against a static exact-source build.

The header marks the data pointer `_Nullable` and the manual page's `ZIP_ER_INVAL` list does not mention the pair, so nothing in the contract tells a caller this is disallowed.

No regress case is added because the harness drives archives through `ziptool`, and there is no way to pass a NULL pointer that way. All 190 ctest tests pass with the guard, and the file is clean against `.clang-format`.


dillof:
Good catch, thanks.

PR #579, opened 2026-09-14
Preserve buffer-source data during cloned writes

seonglae:
Buffer sources implement `ZIP_SOURCE_BEGIN_WRITE_CLONING` by sharing fragment data between the original and output buffers, and a caller can `zip_source_seek_write()` into that shared prefix. `buffer_write()` writes through the shared pointer, modifying the original before commit, so rollback cannot restore it. Because `zip_source_buffer_create()` takes a `const void *`, the same write can fault when the original bytes are in read-only storage.

Copy the still-shared fragment suffix before the first nonempty write into it, from the end toward the write position so a shared prefix survives each allocation and the ownership boundary stays valid if one fails. Appends beyond the cloned prefix and zero-length writes are unchanged.

The regression uses a three-fragment caller-owned source, overwrites four bytes across fragment boundaries, rolls back and checks the caller buffers and reopened source are intact, then repeats, commits, and verifies.

Validation on ARM64 macOS:

- the test-only change fails on the exact unpatched base, and the read-only case aborts under AddressSanitizer;
- an ASan+UBSan matrix of 1,248 clone/seek/write/commit/rollback combinations goes from 660 failures to zero;
- the full normal and ASan+UBSan suites each pass 190/190.

This is defensive public-API hardening with a measured write into read-only storage. No attacker-controlled archive or remote trigger was demonstrated.


dillof:
Good catch, thanks.

We've implemented it more efficiently by making buffers more flexible and only cloning the segments written into.

PR #581, opened 2026-09-18
ci: pin floating branch refs for GitHub Actions

mcc0nnell:
## Summary

- pin the OSS-Fuzz CIFuzz build and run actions to the current commit behind `master`
- pin `lukka/get-cmake` to the current commit behind `latest`
- retain the branch names as comments to make future updates straightforward

## Why

These workflows execute third-party GitHub Actions from floating branch refs:

- `google/oss-fuzz/...@master`
- `lukka/get-cmake@latest`

Pinning them to immutable commit SHAs prevents those action implementations from changing independently of this repository.

The pinned SHAs matched the referenced branch heads when this change was prepared:

- OSS-Fuzz `master`: `0ddfba395c634ac915c7a1315e8c4d6f2bd03d8c`
- get-cmake `latest`: `fffaaafeea488556c2c12dad60690008bc1caacb`

No workflow steps, inputs, or conditions are otherwise changed.

## Validation

- `git diff --check`: PASS
- `actionlint`: PASS for the modified workflows, excluding the repository's existing `runner.workspace` schema warnings
- AppVeyor is currently failing on both this PR head and the base commit; this change only modifies GitHub Actions `uses:` refs


dillof:
We only use CI builds to find errors, we don't use the compiled results, so they can't be used for supply chain attacks.

Therefore, we'd rather automatically get fixes committed to the used actions than exact reproducibility.

